Issue #5, opened 2020-09-30
Simple k-fold cross validation


VolodymyrOrlov:
[K-fold cross validation](https://machinelearningmastery.com/k-fold-cross-validation/) (CV) is a preferred way to evaluate performance of a statistical model. CV is better than just splitting dataset into training/test sets because we use as many data samples for validation as we can get from a single dataset, thus improving estimate of out-of-the-box error.  

SmartCore does not has a method for CV and this is a shame, because any good ML framework must have it.

I think we could start from a simple replica of the Scikit's [sklearn.model_selection.KFold](https://scikit-learn.org/stable/modules/generated/sklearn.model_selection.KFold.html#sklearn.model_selection.KFold). Later on we can add replica of [StratifiedKFold](https://scikit-learn.org/stable/modules/generated/sklearn.model_selection.StratifiedKFold.html#sklearn.model_selection.StratifiedKFold).

If you are not familiar with CV I would start from reading about it [here](https://machinelearningmastery.com/k-fold-cross-validation/) and [here](https://machinelearningmastery.com/repeated-k-fold-cross-validation-with-python/). Next I would look at [Scikit's implementation](https://github.com/scikit-learn/scikit-learn/blob/master/sklearn/model_selection/_split.py) and design a function or a class that does the same for SmartCore. 

We do not have to reproduce class KFold exactly, one way to do it is to write an iterator that spits out K pairs of (train, test) sets. Also, it might be helpful to see [how train/test split is implemented](https://docs.rs/smartcore/0.1.0/smartcore/model_selection/fn.train_test_split.html) in SmartCore

Mec-iS:
Drop here some reference notes:

### procedure
Quoting [here](https://machinelearningmastery.com/k-fold-cross-validation/):
>The general procedure is as follows:
>1. Shuffle the dataset randomly.
>2. Split the dataset into k groups
>3. For each unique group:
    1. Take the group as a hold out or test data set
    2. Take the remaining groups as a training data set
    3. Fit a model on the training set and evaluate it on the test set
    4. Retain the evaluation score and discard the model
>4. Summarize the skill of the model using the sample of model evaluation scores

### stress points
> It is also important that any preparation of the data prior to fitting the model occur on the CV-assigned training dataset within the loop rather than on the broader data set. This also applies to any tuning of hyperparameters.

### example
Having 6 observations split in 3 folds (k-3 CV):
> Three models are trained and evaluated with each fold given a chance to be the held out test set.
[...]

PR #8, opened 2020-10-07
KFold cross-validation

Mec-iS:
#5 

...
* implement `test_indices`, `test_matrices`, `split` to accept a matrix and return a `Vec`
* improve allocation (?)
* make all three to return an iterator (?)

### DIFF
--- a/src/model_selection/mod.rs
+++ b/src/model_selection/mod.rs
@@ -141,7 +141,7 @@ impl BaseKFold for KFold {
 
         let mut return_values: Vec<Vec<usize>> = Vec::new();
         let mut current: usize = 0;
-        for fold_size in fold_sizes {
+        for fold_size in fold_sizes.drain(..) {
             let stop = current + fold_size;
             println!("current, stop {:?}, {:?}", &current, &stop);
             return_values.push(
@@ -155,7 +155,7 @@ impl BaseKFold for KFold {
 
     }
 
-    // fn test_matrices<T: RealNumber, M: Matrix<T>>(&self, X: &M) -> Vec<&M> {
+    // fn test_matrices<T: RealNumber, M: Matrix<T>>(&self, x: &M) -> Vec<&M> {
     //     // Python implementation
     //     // for test_index in self._iter_test_indices(X, y, groups):
     //     //     test_mask = np.zeros(_num_samples(X), dtype=bool)
--- a/src/model_selection/mod.rs
+++ b/src/model_selection/mod.rs
@@ -110,30 +110,24 @@ impl BaseKFold for KFold {
     fn test_indices<T: RealNumber, M: Matrix<T>>(&self, x: &M) -> Vec<Vec<usize>> {
         // number of samples (rows) in the matrix
         let n_samples: usize = x.shape().1;
-        println!("n {:?}", &n_samples);
 
         // initialise indices
         let indices: Vec<usize> = (0..n_samples).collect();
-        println!("starting indices{:?}", &indices);
 
         //  return a new array of given shape n_split, filled with each element of n_samples divided by n_splits.
         let mut fold_sizes = vec![n_samples / self.n_splits as usize; self.n_splits as usize];
-        println!("fold size {:?}", &fold_sizes);
 
         // increment by one if odd
         for i in 0..(n_samples % self.n_splits as usize) {
             fold_sizes[i] = fold_sizes[i] + 1;
         }
-        println!("fold size after correction{:?}", &fold_sizes);
 
         // generate the right array of arrays for test indices
         let mut return_values: Vec<Vec<usize>> = Vec::with_capacity(self.n_splits as usize);
         let mut current: usize = 0;
         for fold_size in fold_sizes.drain(..) {
             let stop = current + fold_size;
-            println!("current, stop {:?}, {:?}", &current, &stop);
             return_values.push(indices[current..stop].to_vec());
-            println!("loop {:?}", &return_values);
             current = stop
         }
 
@@ -157,7 +151,6 @@ impl BaseKFold for KFold {
     fn split<T: RealNumber, M: Matrix<T>>(&self, x: &M) -> Vec<(Vec<usize>, Vec<usize>)> {
         let n_samples: usize = x.shape().1;
         let indices: Vec<usize> = (0..n_samples).collect();
-        println!("starting indices{:?}", &indices);
 
         let mut return_values: Vec<(Vec<usize>, Vec<usize>)> =
             Vec::with_capacity(self.n_splits as usize); // TODO: init nested vecs with capacities by getting the length of test_index vecs
@@ -214,7 +207,6 @@ mod tests {
         let x: DenseMatrix<f64> = DenseMatrix::rand(100, 33);
         let test_indices = k.test_indices(&x);
 
-        println!("{:?}", &test_indices);
         assert_eq!(test_indices[0], (0..11).collect::<Vec<usize>>());
         assert_eq!(test_indices[1], (11..22).collect::<Vec<usize>>());
         assert_eq!(test_indices[2], (22..33).collect::<Vec<usize>>());
@@ -226,7 +218,6 @@ mod tests {
         let x: DenseMatrix<f64> = DenseMatrix::rand(100, 34);
         let test_indices = k.test_indices(&x);
 
-        println!("{:?}", &test_indices);
         assert_eq!(test_indices[0], (0..12).collect::<Vec<usize>>());
         assert_eq!(test_indices[1], (12..23).collect::<Vec<usize>>());
         assert_eq!(test_indices[2], (23..34).collect::<Vec<usize>>());
@@ -239,6 +230,7 @@ mod tests {
         let test_masks = k.test_masks(&x);
 
         for t in &test_masks[0][0..11] {
+            // TODO: this can be prob done better
             assert_eq!(*t, true)
         }
         for t in &test_masks[0][11..22] {
@@ -258,7 +250,6 @@ mod tests {
         let k = KFold { n_splits: 2 };
         let x: DenseMatrix<f64> = DenseMatrix::rand(100, 22);
         let train_test_splits = k.split(&x);
-        print!("{:?}", &train_test_splits);
 
         assert_eq!(train_test_splits[0].1, (0..11).collect::<Vec<usize>>());
         assert_eq!(train_test_splits[0].0, (11..22).collect::<Vec<usize>>());
--- a/src/model_selection/mod.rs
+++ b/src/model_selection/mod.rs
@@ -138,7 +138,7 @@ trait BaseKFold {
 
 }
 
-struct KFold {
+pub struct KFold {
     n_splits: i32,
     shuffle: bool,
     random_state: i32, 
--- a/src/model_selection/mod.rs
+++ b/src/model_selection/mod.rs
@@ -105,9 +105,9 @@ trait BaseKFold {
     /// Returns masksk corresponding to test sets
     fn test_masks<T: RealNumber, M: Matrix<T>>(&self, x: &M) -> Vec<Vec<bool>>;
 
-    // /// Return a tuple containing the the training set indices for that split and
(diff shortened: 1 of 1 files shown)


src/model_selection/mod.rs
```diff
@@ -107,3 +107,43 @@ mod tests {
         assert_eq!(x_test.shape().0, y_test.len());
     }
 }
+
+
+///
+/// KFold Cross-Validation
+///
+/// Entities involved in the KFold procedure:
```
VolodymyrOrlov, line 115: Will our user benefit from knowing the list of entities in the KFold procedure? 

Doc strings that start from `///` will be displayed in https://docs.rs/smartcore/0.1.0/smartcore/model_selection/index.html and it is important to keep description of the methods and classes short but informative. The danger of keeping a lot of unnecessary information here is that this information might get out-of-sync if not properly maintained. Doc strings should describe the function of the method or a class, meaning of all parameters, some technical details like errors and default values and a short example. Any other information that describes algorithm in greater details should go into the user's manual: https://github.com/smartcorelib/smartcorelib.org

src/model_selection/mod.rs
```diff
@@ -107,3 +107,43 @@ mod tests {
...
+///     * a dataset
+///     * a number k of groups (k-folds)
+/// 
+/// Procedure in `cross_validate()`:
```
VolodymyrOrlov, line 119: user's manual: https://github.com/smartcorelib/smartcorelib.org is a better place for the algorithm's pseudocode.

VolodymyrOrlov (changes requested):
Hi @Mec-iS, this is a great start! I have couple of suggestions for you that will help to improve this PR:

* Run `cargo fmt` every time before committing your code, otherwise automatic build will fail if code is incorrectly formatted. 
* Test how project's documentation looks by running `cargo doc --no-deps --features "nalgebra-bindings ndarray-bindings"`. This command will generate html with documentation in the `target/doc/` folder. 

I am looking forward to see implementation of the function that splits dataset into k folds!

Mec-iS:
Thanks. As a next step I will try to implement the interface with only the matrix as input (not considering shuffle and groups).

Mec-iS:
@VolodymyrOrlov can you specify the meaning of [this assignment](https://github.com/scikit-learn/scikit-learn/blob/0864c5896804bc1066c02bcb1443c962cabdc420/sklearn/model_selection/_split.py#L439)

VolodymyrOrlov:
I think the idea is to make sure that total number of samples from all folds equals total number of samples. Here is an example
image
without this line we have only 2 + 2 + 2 = 6 samples in all folds. If you uncomment this line and run the code again total number of samples equal number of samples in X, 3 + 2 + 2 = 7:
image

Mec-iS:
Please check that `test_indices` works as expected

Mec-iS:
Ready for review :+1:

src/model_selection/mod.rs
```diff
@@ -81,6 +81,100 @@ pub fn train_test_split<T: RealNumber, M: Matrix<T>>(
     (x_train, x_test, y_train, y_test)
 }
 
+///
+/// KFold Cross-Validation
+///
+trait BaseKFold {
```
VolodymyrOrlov, line 87: \+ pub, otherwise the trait `BaseKFold` is invisible outside of this module
```suggestion
pub trait BaseKFold {
```

src/model_selection/mod.rs
```diff
@@ -81,6 +81,100 @@ pub fn train_test_split<T: RealNumber, M: Matrix<T>>(
...
+    /// Return a tuple containing the the training set indices for that split and
+    /// the testing set indices for that split.
+    fn split<T: RealNumber, M: Matrix<T>>(&self, x: &M) -> Vec<(Vec<usize>, Vec<usize>)>;
+}
+
+///
+/// An implementation of KFold
+///
+pub struct KFold {
+    n_splits: i32, // cannot exceed std::usize::MAX
```
VolodymyrOrlov, line 103: \+ pub, unless you want to add a public method `KFold::new` that will create an instance of this class

```suggestion
    pub n_splits: i32, // cannot exceed std::usize::MAX
```

src/model_selection/mod.rs
```diff
@@ -81,6 +81,100 @@ pub fn train_test_split<T: RealNumber, M: Matrix<T>>(
...
+            // set mask's indices to true according to test indices
+            for i in test_index {
+                test_mask[i] = true; // can be implemented with map()
+            }
+            return_values.push(test_mask);
+        }
+        return return_values;
+    }
+
+    fn split<T: RealNumber, M: Matrix<T>>(&self, x: &M) -> Vec<(Vec<usize>, Vec<usize>)> {
```
VolodymyrOrlov, line 151: Looks like this method does not split **x** in the same way as Scikit  does. For example, 3-fold split of 3x4 matrix in Python would look like this:
image. 

While our implementation outputs this:
image
image

I suggest this test to verify parity between Scikit learn implementation and ours:
```suggestion
#[test]
    fn numpy_parity_test() {
        let k = KFold { n_splits: 3 };
        let x: DenseMatrix<f64> = DenseMatrix::rand(10, 4);
        let expected: Vec<(Vec<usize>, Vec<usize>)> = vec!(
            (vec!(4, 5, 6, 7, 8, 9), vec!(0, 1, 2, 3)),
            (vec!(0, 1, 2, 3, 7, 8, 9), vec!(4, 5, 6)),
            (vec!(0, 1, 2, 3, 4, 5, 6), vec!(7, 8, 9))
        );
        for ((train, test), (expected_train, expected_test)) in k.split(&x).into_iter().zip(expected) {
            assert_eq!(test, expected_test);
            assert_eq!(train, expected_train);
        }        
    }
```

VolodymyrOrlov (changes requested):

VolodymyrOrlov:
@Mec-iS the code looks good! Could of things left before I can proceed to merging this PR:

1. Looks like our implementation does not split data the same way as Scikit-learn's does. 
1. Arguably flag `shuffle` is the most important feature of K-fold split. Let's add it and set it to true by default.

Mec-iS:
I had problem understanding the shape of the matrix, in my understanding:
* `M::shape().0` returns the dimensions (columns)
* `M::shape().1` returns the samples (rows)
Please provide guidance.

Mec-iS:
ahah, sorry I got the dimensions confused. I just had to switch which index in `shapes` I was using and now the tests seem to pass.

src/model_selection/mod.rs
Mec-iS: I don't see any usage pattern that requires `n_splits` to be accessible outside of the type.

Mec-iS:
Question: in my [last commit I added this line](https://github.com/smartcorelib/smartcore/pull/8/files#diff-ca3045cade7b17a743956d16a5d135deR153) to explicitly define `return_values` as immutable before returning ownership, the same for `indices` in `test_indices` after the optional shuffle; is this worth it or the compiler can do this kind of optimisation automatically? I couldn't find a reference to this in Rust doc so I am not sure. If not, let me know I will revert the commit.

VolodymyrOrlov:
> I had problem understanding the shape of the matrix, in my understanding:
> 
> * `M::shape().0` returns the ~dimensions (columns)~ the samples
> * `M::shape().1` returns the ~samples (rows)~ the dimensions
> 
> Please provide guidance. I couldn't find this information clearly stated in the documentation.

Good catch, thanks for pointing it out! I'll fix documentation of the `shape` method

VolodymyrOrlov:
> Question: in my [last commit I added this line](https://github.com/smartcorelib/smartcore/pull/8/files#diff-ca3045cade7b17a743956d16a5d135deR153) to explicitly define `return_values` as immutable before returning ownership, the same for `indices` in `test_indices` after the optional shuffle; is this worth it or the compiler can do this kind of optimisation automatically? I couldn't find a reference to this in Rust doc so I am not sure. If not, let me know I will revert the commit.

No, it is not necessary since you pass ownership of the `return_values` to the caller. A function that calls the method `test_indices` will be able to modify `return_values` and I see no reason why we should not let it to do so. 

I am sure you've read this book but will leave this link here in case you haven't: https://doc.rust-lang.org/book/ch04-00-understanding-ownership.html.

src/model_selection/mod.rs
```diff
@@ -81,6 +83,118 @@ pub fn train_test_split<T: RealNumber, M: Matrix<T>>(
...
+                .collect::<Vec<usize>>(); // filter train indices out according to mask
+            let test_index = indices
+                .iter()
+                .enumerate()
+                .filter(|&(idx, _)| test_index[idx] == true)
+                .map(|(idx, _)| idx)
+                .collect::<Vec<usize>>(); // filter tests indices out according to mask
+            return_values.push((train_index, test_index))
+        }
+        return return_values;
```
VolodymyrOrlov, line 194: nit: you can simplify this to
```suggestion
        return_values
```

src/model_selection/mod.rs
```diff
@@ -81,6 +83,118 @@ pub fn train_test_split<T: RealNumber, M: Matrix<T>>(
...
+        let mut return_values: Vec<Vec<usize>> = Vec::with_capacity(self.n_splits as usize);
+        let mut current: usize = 0;
+        for fold_size in fold_sizes.drain(..) {
+            let stop = current + fold_size;
+            return_values.push(indices[current..stop].to_vec());
+            current = stop
+        }
+
+        let return_values = return_values;
+        return return_values;
```
VolodymyrOrlov, line 154: nit: you can simplify this to
```suggestion
        return_values
```

src/model_selection/mod.rs
```diff
@@ -81,6 +83,118 @@ pub fn train_test_split<T: RealNumber, M: Matrix<T>>(
...
+        for test_index in self.test_indices(x).drain(..) {
+            // init mask
+            let mut test_mask = vec![false; x.shape().0];
+            // set mask's indices to true according to test indices
+            for i in test_index {
+                test_mask[i] = true; // can be implemented with map()
+            }
+            return_values.push(test_mask);
+        }
+        return return_values;
```
VolodymyrOrlov, line 168: nit: you can simplify this to
```suggestion
        return_values
```

Mec-iS:
> since you pass ownership of the return_values to the caller. A function that calls the method test_indices will be able to modify return_values and I see no reason why we should not let it to do so.

yes but I dont see the necessity of passing the ownership of a mutable vector. It could be an interface leak to allow mutability on something that is not meant to be modified. Once the indices and the splits are defined there is no reason for the developer-user to be able to modify it. That is why is conceptually more correct to return an immutable maybe.

VolodymyrOrlov:
> > since you pass ownership of the return_values to the caller. A function that calls the method test_indices will be able to modify return_values and I see no reason why we should not let it to do so.
> 
> yes but I dont see the necessity of passing the ownership of a mutable vector. It could be an interface leak to allow mutability on something that is not meant to be modified. Once the indices and the splits are defined, there is no reason for the developer-user to be able to modify it. That is why is conceptually more correct to return an immutable maybe.

I agree with you conceptually but I don't know an easy way to do it without wrapping your return value into another struct that will prevent a caller from modifying it. https://users.rust-lang.org/t/how-to-return-an-immutable-reference-of-a-mutable-vector/30773/5

I think wrapping return value into protective shell is an overkill

Mec-iS:
>I agree with you conceptually but I don't know an easy way to do it without wrapping your return value into another struct that will prevent a caller from modifying it. https://users.rust-lang.org/t/how-to-return-an-immutable-reference-of-a-mutable-vector/30773/5
I think wrapping return value into protective shell is an overkill

no wait, something missing. I have already implemented it with re-assigning the return value to an immutable variable: `let return_values = return_values`, the right side is a mutable vector but once re-assigned to an immutable variable it gets immutable and later returned by the next line (`return_values`). According to my current knowledge of Rust it should work, I just wanted comfirmation that this makes sense (the compiler does not complain but I didnt check what happens in memory)

VolodymyrOrlov (approved):
lgtm

Issue #9, opened 2020-10-12
SVM


VolodymyrOrlov:
Implement Support vector machine (SVM) classifier and regressor.

The theory behind SVM is described in the book ["An Introduction to Statistical Learning" by Gareth James, Daniela Witten, Trevor Hastie and Robert Tibshirani](http://faculty.marshall.usc.edu/gareth-james/ISL/). 

[This paper](https://www.jmlr.org/papers/volume6/fan05a/fan05a.pdf) (+ references ) describes one approach to implementing Sequential minimal optimization (SMO) algorithm that can be used to speedup SVR training, while library (and paper) [LaSVM](https://leon.bottou.org/projects/lasvm) describes fast implementation of another algorithm for SVC. 

SmartCore should support at least these 3 kernels:

* linear
* [polynomial](https://en.wikipedia.org/wiki/Polynomial_kernel)
* [rbf](https://en.wikipedia.org/wiki/Radial_basis_function_kernel)

VolodymyrOrlov:
SVC and SVR implementations are not in development

PR #12, opened 2020-10-15
epsilon-SVR + new methods in BaseMatrix and BaseVector

VolodymyrOrlov:
This PR covers part of #9 

Here I introduce:
* [e-SVR](https://scikit-learn.org/stable/modules/generated/sklearn.svm.SVR.html#sklearn.svm.SVR) implementation
* 3 new methods in `linalg` module: `BaseVector::dot`, `BaseVector::approximate_eq`, `BaseMatrix::get_row`

SVR implementation lacks proper documentation but I plan to add it in a separate PR

### DIFF
--- a/src/svm/svr.rs
+++ b/src/svm/svr.rs
@@ -1,0 +1,538 @@
+//! # Epsilon-Support Vector Regression.
+//!
+//! Example
+//!
+//! ```
+//! use smartcore::linalg::naive::dense_matrix::*;
+//! use smartcore::linear::linear_regression::*;
+//! use smartcore::svm::*;
+//! use smartcore::svm::svr::{SVR, SVRParameters};
+//!
+//! // Longley dataset (https://www.statsmodels.org/stable/datasets/generated/longley.html)
+//! let x = DenseMatrix::from_2d_array(&[
+//!               &[234.289, 235.6, 159.0, 107.608, 1947., 60.323],
+//!               &[259.426, 232.5, 145.6, 108.632, 1948., 61.122],
+//!               &[258.054, 368.2, 161.6, 109.773, 1949., 60.171],
+//!               &[284.599, 335.1, 165.0, 110.929, 1950., 61.187],
+//!               &[328.975, 209.9, 309.9, 112.075, 1951., 63.221],
+//!               &[346.999, 193.2, 359.4, 113.270, 1952., 63.639],
+//!               &[365.385, 187.0, 354.7, 115.094, 1953., 64.989],
+//!               &[363.112, 357.8, 335.0, 116.219, 1954., 63.761],
+//!               &[397.469, 290.4, 304.8, 117.388, 1955., 66.019],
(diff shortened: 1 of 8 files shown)


vadimzaliva (dismissed):

src/svm/svr.rs
```diff
@@ -0,0 +1,538 @@
...
+            x: x,
+            grad: [eps + y, eps - y],
+            k: k_v,
+            alpha: [T::zero(), T::zero()],
+        }
+    }
+}
+
+impl<'a, T: RealNumber, M: Matrix<T>, K: Kernel<T, M::RowVector>> Optimizer<'a, T, M, K> {
+    fn optimize(
```
Mec-iS, line 212: it would be helpful a docstring here

src/svm/svr.rs
```diff
@@ -0,0 +1,538 @@
...
+            }
+            if g > self.gmax && a > T::zero() {
+                self.gmax = g;
+                self.gmaxindex = 1;
+                self.svmax = i;
+            }
+        }
+    }
+
+    fn smo(mut self) -> (Vec<M::RowVector>, Vec<T>, T) {
```
Mec-iS, line 281: it would be helpful to have a docstring here

src/svm/svr.rs
```diff
@@ -0,0 +1,538 @@
...
+            SVRParameters {
+                eps: 2.0,
+                c: 10.0,
+                tol: 1e-3,
+            },
+        )
+        .and_then(|lr| lr.predict(&x))
+        .unwrap();
+
+        println!("{:?}", y_hat);
```
Mec-iS, line 500: do we want to have print statements in tests?

src/svm/svr.rs
```diff
@@ -0,0 +1,538 @@
...
+                    .map(|vi| self.kernel.apply(&self.sv[v2].x, &vi.x))
+                    .collect()
+            });
+
+            let mut curv = self.sv[v1].k + self.sv[v2].k - T::two() * k1[self.sv[v2].index];
+            if curv <= T::zero() {
+                curv = self.tau;
+            }
+
+            if i != j {
```
Mec-iS, line 350: I dont know if it is applicable in this particular case, but I believe that the Rust way of doing this is patter matching (`match` `case`)

VolodymyrOrlov: Usually I use `match` when I have 2 or more conditions to go through. In this case I have a simple if/else case that doesn't justify matching.

src/svm/svr.rs
VolodymyrOrlov: thanks for spotting it!

src/svm/svr.rs
VolodymyrOrlov: I've added references and documented this method a bit. I plan to add more solid description of SVR once I am done with SVC. It will probably be done in a separate PR.

Mec-iS (approved):

Issue #14, opened 2020-10-16
Naive Bayes (NB) Classifier


VolodymyrOrlov:
Implement Base NB classifier that doesn't make any assumptions about the underlying distribution of **x**. 

https://scikit-learn.org/stable/modules/naive_bayes.html

We need something like this (pseudocode):

```
trait NBDistribution:
    
    // Fit distribution to some continuous or discrete data
    def fit(x: Matrix<T>) -> NBDistribution
    
    // prior of class k 
    def prior(k) -> T

    // conditional probability of feature j give class k
    def conditional_probability(k, j)-> T

class BaseNaiveBayes:
    
    // "Fits" NB. This method validates and remembers parameters
    def fit(distribution: NBDistribution)
    
    // Calculates likelihood of labels using stored probabilities and X. Returns vector with estimated labels
    def predict(x: Matrix<T>) -> Vector<T>
```

Once we have `BaseNaiveBayes` we can implement Gaussian Naive Bayes, Multinomial Naive Bayes and Bernoulli Naive Bayes as concrete implementations of trait `NBDistribution`

morenol:
Remaining NB classifiers to implement:

-[ ] [Multinomial Naive Bayes](https://scikit-learn.org/stable/modules/naive_bayes.html#multinomial-naive-bayes)
-[ ] [Complement Naive Bayes](https://scikit-learn.org/stable/modules/naive_bayes.html#complement-naive-bayes)
-[ ] [Bernoulli Naive Bayes](https://scikit-learn.org/stable/modules/naive_bayes.html#bernoulli-naive-bayes)

And also a partial_fit()  for the cases where it makes sense, right? @VolodymyrOrlov

VolodymyrOrlov:
@morenol can I close this issue, or would you like to keep it open until `Complement Naive Bayes` is implemented?

morenol:
I think that we should close this when the documentation is already in the code.

PR #15, opened 2020-10-20
feat: add Naive Bayes and CategoricalNB 

morenol:
See https://github.com/smartcorelib/smartcore/issues/14 for details


src/naive_bayes/mod.rs
```diff
@@ -0,0 +1,22 @@
...
+
+    fn prior<T: RealNumber>(k: T) -> T;
+
+    fn conditional_probability<T: RealNumber, M: Matrix<T>>(k: T, j: M::RowVector) -> T;
+}
+
+pub struct BaseNaiveBayes;
+
+impl BaseNaiveBayes {
+    pub fn fit<D: NBDistribution>(&mut self, distribution: D) -> Result<(), Failed> {
```
VolodymyrOrlov, line 15: Shell we return an instance of `BaseNaiveBayes` here? Otherwise, what user will get after fitting `BaseNaiveBayes ` to data? Also, method `fit` should be static.
```suggestion
    pub fn fit<D: NBDistribution>(distribution: D) -> Result<Self, Failed> {
```

src/naive_bayes/mod.rs
```diff
@@ -0,0 +1,22 @@
...
+        unimplemented!()
+    }
+
+    pub fn predict<T: RealNumber, M: Matrix<T>>(x: &M) -> Result<M::RowVector, Failed> {
```
VolodymyrOrlov, line 19: you might have forgot `self` here
```suggestion
    pub fn predict<T: RealNumber, M: Matrix<T>>(&self, x: &M) -> Result<M::RowVector, Failed> {
```

src/naive_bayes/mod.rs
```diff
@@ -0,0 +1,22 @@
+use crate::error::Failed;
+use crate::linalg::Matrix;
+use crate::math::num::RealNumber;
+pub trait NBDistribution {
+    fn fit<T: RealNumber, M: Matrix<T>>(&self, x: &M, y: M::RowVector) -> Self;
```
VolodymyrOrlov, line 5: Let's make it static
```suggestion
    fn fit<T: RealNumber, M: Matrix<T>>(x: &M, y: M::RowVector) -> Self;
```

src/naive_bayes/mod.rs
```diff
@@ -0,0 +1,22 @@
...
+
+    fn prior<T: RealNumber>(k: T) -> T;
```
VolodymyrOrlov, line 7: \+ self
```suggestion
    fn prior<T: RealNumber>(&self, k: T) -> T;
```

src/naive_bayes/mod.rs
```diff
@@ -0,0 +1,22 @@
...
+
+    fn conditional_probability<T: RealNumber, M: Matrix<T>>(k: T, j: M::RowVector) -> T;
```
VolodymyrOrlov, line 9: \+ self
```suggestion
    fn conditional_probability<T: RealNumber, M: Matrix<T>>(&self, k: T, j: M::RowVector) -> T;
```

src/naive_bayes/mod.rs
```diff
@@ -0,0 +1,51 @@
...
+    fn fit<T: RealNumber, M: Matrix<T>>(x: &M, y: &M::RowVector) -> Self;
+
+    // Prior of class k
+    fn prior<T: RealNumber>(&self, k: T) -> T;
+
+    // Conditional probability of feature j give class k
+    fn conditional_probability<T: RealNumber, M: Matrix<T>>(&self, k: T, j: &M::RowVector) -> T;
+
+    // Possible classes of the distribution
+    fn classes<T: RealNumber, M: Matrix<T>>(&self) -> M::RowVector;
```
VolodymyrOrlov, line 16: I think this function should return simple Vec<T> since it represents an unordered list of classes that doesn't benefit from Vector methods like dot product, norm, e.t.c.

src/naive_bayes/mod.rs
```diff
@@ -0,0 +1,51 @@
...
+pub struct BaseNaiveBayes<D: NBDistribution> {
+    distribution: D,
+}
+
+impl<D: NBDistribution> BaseNaiveBayes<D> {
+    pub fn fit(distribution: D) -> Result<Self, Failed> {
+        Ok(Self { distribution })
+    }
+
+    pub fn predict<T: RealNumber, M: Matrix<T>>(&self, x: &M) -> Result<Vec<T>, Failed> {
```
VolodymyrOrlov, line 28: Implementation looks good! But we need a simple unit test for this function

src/naive_bayes/mod.rs
```diff
@@ -0,0 +1,186 @@
...
+        y_labels: Vec<T>,
+        label_count: Vec<usize>,
+        data_size: usize,
+        x: M,
+        y: M::RowVector,
+    }
+
+    impl<T: RealNumber, M: Matrix<T>> NBDistribution<T, M> for FakeDistribution<T, M> {
+        fn fit(x: &M, y: &M::RowVector) -> Self {
+            let x = x.clone();
```
VolodymyrOrlov, line 87: for the MultinomialNBDistribution let's make sure we don't do excessive cloning. For example, [Multinomial distribution is parametrized by parameter _theta_](https://scikit-learn.org/stable/modules/naive_bayes.html#multinomial-naive-bayes) hence all we need to successfully execute methods _prior_, _conditional_probability_ and _classes_ are thetas, class probabilities and class values. We do not need to keep _x_ and _y_ in the struct that represents the distribution.

VolodymyrOrlov (changes requested):

VolodymyrOrlov:
@morenol is this PR still a draft or would you like me to take another look at it?

src/naive_bayes/mod.rs
```diff
@@ -0,0 +1,239 @@
...
+use crate::linalg::Matrix;
+use crate::math::num::RealNumber;
+use std::marker::PhantomData;
+
+/// Distribution used in the Naive Bayes classifier.
+pub trait NBDistribution<T: RealNumber, M: Matrix<T>> {
+    /// Prior of class k
+    fn prior(&self, k: T) -> T;
+
+    /// Conditional probability of feature j given class k
```
VolodymyrOrlov, line 12: I think doc string for this function should be
```
Conditional probability of sample j given class k
```
because this function returns product of conditional probabilities, right?
image

src/naive_bayes/mod.rs
```diff
@@ -0,0 +1,239 @@
+use crate::error::Failed;
+use crate::linalg::BaseVector;
+use crate::linalg::Matrix;
+use crate::math::num::RealNumber;
+use std::marker::PhantomData;
+
+/// Distribution used in the Naive Bayes classifier.
+pub trait NBDistribution<T: RealNumber, M: Matrix<T>> {
```
VolodymyrOrlov, line 8: Let's make this trait pub(crate). This trait and its implementations are not useful outside of NB and  I think we can simplify API with another class that is similar to Scikit learn's CategoricalNB. This new class should combine distribution with `BaseNaiveBayes ` to make sure we can do
```
let nb = BaseNaiveBayes::fit(&x, &y, Default::default()).unwrap();
```
instead of
```
let distribution = CategoricalNB::<f64>::fit(&x, &y, alpha).unwrap();
let nb = BaseNaiveBayes::fit(distribution).unwrap();
```

src/naive_bayes/mod.rs
```diff
@@ -0,0 +1,239 @@
...
+        let mut y_hat = M::RowVector::zeros(rows);
+        for (i, prediction) in predictions.iter().enumerate().take(rows) {
+            y_hat.set(i, *prediction);
+        }
+        Ok(y_hat)
+    }
+}
+
+/// Naive Bayes classifier for categorical features
+pub struct CategoricalNB<T: RealNumber> {
```
VolodymyrOrlov, line 69: As I suggested in my earlier comment let's rename this class to CategoricalNBDistribution and make it `pub (crate) ` or private

src/naive_bayes/mod.rs
```diff
@@ -0,0 +1,239 @@
...
+        &self.class_labels
+    }
+}
+
+impl<T: RealNumber> CategoricalNB<T> {
+    /// Fits the distribution to a NxM matrix where N is number of samples and M is number of features.
+    /// * `x` - training data.
+    /// * `y` - vector with target values (classes) of length N.
+    /// * `alpha` - Additive (Laplace/Lidstone) smoothing parameter (0 for no smoothing).
+    pub fn fit<M: Matrix<T>>(x: &M, y: &M::RowVector, alpha: T) -> Result<Self, Failed> {
```
VolodymyrOrlov, line 114: Let's make sure that alpha is an optional parameter. By default it should be 1.0.

src/naive_bayes/mod.rs
```diff
@@ -0,0 +1,239 @@
...
+pub struct CategoricalNB<T: RealNumber> {
+    class_labels: Vec<T>,
+    class_probabilities: Vec<T>,
+    coef: Vec<Vec<Vec<T>>>,
+    feature_categories: Vec<Vec<T>>,
+}
+
+impl<T: RealNumber, M: Matrix<T>> NBDistribution<T, M> for CategoricalNB<T> {
+    fn prior(&self, k: T) -> T {
+        match self.class_labels.iter().position(|&t| t == k) {
```
VolodymyrOrlov, line 78: I feel that there should be a way to avoid searching for a position in array where t == k. While we can assume that number of classes is small, usually < 100 it is better to avoid inefficiencies if possible.  
One way to make sure we do not search array every time when we need to get a prior is to store all distinct class labels in a vector and use position of the class in a vector as an ID for this class. This way you know that you class labels starts from 0 and go up to k, where k is number of classes. In this case your priors will be stored in a vector of length k with position of each element corresponding to a prior of class your class.
To go back to "real" class values all you need is to replace class IDs with real class value from the stored vector.

src/naive_bayes/mod.rs
```diff
@@ -0,0 +1,239 @@
...
+                current_count += T::one();
+            } else {
+                classes_count.push(current_count);
+                class_labels.push(y_sorted[idx]);
+                current_count = T::one()
+            }
+            classes_count.push(current_count);
+        }
+
+        let x = x.transpose();
```
VolodymyrOrlov, line 155: Why do you need to do transpose?

VolodymyrOrlov (changes requested):

src/naive_bayes/mod.rs
morenol: Do you mean, prior should receive as an input an index instead of the actual value of the label?

VolodymyrOrlov: yes for this method and for `conditional_probability`.

src/naive_bayes/mod.rs
VolodymyrOrlov: Sorry for confusion. First code snippet should be:
```
let nb = CategoricalNB::fit(&x, &y, Default::default()).unwrap();
```

src/naive_bayes/mod.rs
morenol: not needed, I can do the same with get_col(). Removing

src/naive_bayes/mod.rs
morenol: done!

src/naive_bayes/mod.rs
morenol: Done, with that change the compiler warns about unused code, but I guess that it is ok for now and it should be solved once we implement the public API

src/naive_bayes/mod.rs
morenol: implemented in https://github.com/smartcorelib/smartcore/pull/15/commits/456249bdb8ce466edbf5747b21d014cfecccfb38

src/naive_bayes/mod.rs
VolodymyrOrlov: Can you also implement public interface?
The goal is to make sure we can fit categorical NB in one line:

```
let nb = CategoricalNB::fit(&x, &y, Default::default()).unwrap();
```

Also, do you think we should move Categorical NB + corresponding distribution into a separate file? Or you'd prefer to keep it in the _mod.rs_ for now?

morenol: Yes, I can implement that! 

Regarding to the other one, in scikit they have CategoricalNB directly in the naive_bayes module, I think that I can implement them in another file, but I also think that it would be fine to reimport them in the the .mod file, what do you think?

VolodymyrOrlov:
@morenol there is one more optimization we can do that I haven't considered when you asked me to review interface. 

In this formula you are calculating product of conditional probabilities and then multiply it by estimate of class probability:
image
One neat trick you can use to make calculations more stable and fast is to take logarithm of probability estimates and change this formula to 
image

This optimization will require change in the interface NBDistribution. I do not insist on this optimization and can do it myself in a separate PR after this PR is merged.

src/naive_bayes/mod.rs
VolodymyrOrlov: I just feel that when we'll add other types of NB the file _mod.rs_ will get too large and tricky to maintain.

VolodymyrOrlov (approved):
The code looks good! Feel free to merge it at your convenience!

src/naive_bayes/mod.rs
morenol: Implemented in https://github.com/smartcorelib/smartcore/pull/15/commits/7620226e2d27d895749847583b38c52b624ce623

morenol:
> The code looks good! Feel free to merge it at your convenience!
Hi @VolodymyrOrlov 
I think that I don't have rights to do that

VolodymyrOrlov:
> > The code looks good! Feel free to merge it at your convenience!
> 
> Hi @VolodymyrOrlov
> I think that I don't have rights to do that

I've added your GitHub ID to the list of contributors. You should have got an invitation by now. Once you accept it you should be able to merge this PR. Welcome to the team!

Issue #20, opened 2020-11-03
Ridge Regression


VolodymyrOrlov:
Implement Ridge Regression that is similar in functionality to [Scikit's Ridge](https://scikit-learn.org/stable/modules/generated/sklearn.linear_model.Ridge.html#sklearn.linear_model.Ridge). 

We are looking to support following parameters:
* alpha
* solver

The implementation should support at least Cholesky decomposition for `solver`

https://ncss-wpengine.netdna-ssl.com/wp-content/themes/ncss/pdf/Procedures/NCSS/Ridge_Regression.pdf
https://www.cs.ubc.ca/~schmidtm/Courses/540-F14/leastSquares.pdf

PR #21, opened 2020-11-03
feat: adds Cholesky matrix decomposition

VolodymyrOrlov:
This is the first step in implementing Ridge Regression (#20). 

In this PR I add [Cholesky decomposition routine](https://en.wikipedia.org/wiki/Cholesky_decomposition) that can be extended by any matrix that implements `BaseMatrix` trait


Mec-iS (approved):
cool :+1:

PR #25, opened 2020-11-12
Add capability to convert a slice to a BaseVector

morenol:
I wanted to implement a generic way to convert a Vec<T> to a BaseVector<T> but then I realized that I could do that as &[T] which fits in the &Vec<T> case.

I also changed the implementation of RealNumberVector over BaseVector instead of over Vec<T> so I can use that in the implementation of the other Naive Bayes classifiers without the need of creating a Vec<T> from the data. Everything should work at the same way. The only major concern there is that if  BaseVector  is on scope it would be necessary to call that at this way because it conflicts with the BaseVector unique function:

```rust
RealNumber::unique(v);
```

### DIFF
--- a/src/linalg/mod.rs
+++ b/src/linalg/mod.rs
@@ -83,6 +83,21 @@ pub trait BaseVector<T: RealNumber>: Clone + Debug {
         self.len() == 0
     }
 
+    /// Create a new vector from a &[T]
+    /// ```
+    /// use smartcore::linalg::naive::dense_matrix::*;
+    /// let slice: &[f64] = &[0., 0.5, 2., 3., 4.];
+    /// let a: Vec<f64> = BaseVector::from_slice(slice);
+    /// assert_eq!(a, vec![0., 0.5, 2., 3., 4.]);
+    /// ```
+    fn from_slice(f: &[T]) -> Self {
+        let mut v = Self::zeros(f.len());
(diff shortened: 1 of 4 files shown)


VolodymyrOrlov (dismissed):
To be honest, I think I should remove `vector.rs` since BaseVector already has an implementation of the `unique` method, although the return type of the latter function does not exactly match type returned by the function `unique` from the `vectors.rs`.

morenol:
> To be honest, I think I should remove `vector.rs` since BaseVector already has an implementation of the `unique` method, although the return type of the latter function does not exactly match type returned by the function `unique` from the `vectors.rs`.

That's being used in src/metrics/cluster_helpers.rs, maybe we could just rename that to avoid conflicts

VolodymyrOrlov:
> > To be honest, I think I should remove `vector.rs` since BaseVector already has an implementation of the `unique` method, although the return type of the latter function does not exactly match type returned by the function `unique` from the `vectors.rs`.
> 
> That's being used in src/metrics/cluster_helpers.rs, maybe we could just rename that to avoid conflicts

yes, let's rename it to `unique_with_indices` for now. I'll do refactoring and remove it later, some time before releasing 0.2.0

VolodymyrOrlov (approved):
LGTM

PR #27, opened 2020-11-13
Implement GaussianNB

morenol:
Implementation of a Gaussian Naive bayes classifiers similar to https://scikit-learn.org/stable/modules/generated/sklearn.naive_bayes.CategoricalNB.html#sklearn.naive_bayes.GaussianNB

### DIFF
--- a/src/naive_bayes/gaussian.rs
+++ b/src/naive_bayes/gaussian.rs
@@ -10,10 +10,14 @@ use serde::{Deserialize, Serialize};
 /// Naive Bayes classifier for categorical features
 #[derive(Serialize, Deserialize, Debug)]
 struct GaussianNBDistribution<T: RealNumber> {
+    /// class labels known to the classifier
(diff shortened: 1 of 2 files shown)


VolodymyrOrlov:
@morenol is this PR ready for review? I ask because it is still marked as draft.

morenol:
Not yet @VolodymyrOrlov, I think that it works but currently there is an excessive amount of cloning that I want to remove

VolodymyrOrlov:
Can you also compare your coefficients with Scikit's Gaussian NB like we did for Categorical NB? https://scikit-learn.org/stable/modules/naive_bayes.html#gaussian-naive-bayes. It would be great if you can add a unit test that compares coefficients directly. Comparing `y_hat` with `y` might not be precise enough.

VolodymyrOrlov (approved):
lgtm

PR #28, opened 2020-11-14
Use log likelihood to make calculations more stable

morenol:
Changing interface of NBDistribution to use log likelihood in order to make calculations more fast and stable

### DIFF
--- a/src/naive_bayes/categorical.rs
+++ b/src/naive_bayes/categorical.rs
@@ -8,7 +8,7 @@ use serde::{Deserialize, Serialize};
 /// Naive Bayes classifier for categorical features
 struct CategoricalNBDistribution<T: RealNumber> {
     class_labels: Vec<T>,
-    class_probabilities: Vec<T>,
+    class_priors: Vec<T>,
     coef: Vec<Vec<Vec<T>>>,
     feature_categories: Vec<Vec<T>>,
 }
@@ -18,24 +18,24 @@ impl<T: RealNumber, M: Matrix<T>> NBDistribution<T, M> for CategoricalNBDistribu
         if class_index >= self.class_labels.len() {
             T::zero()
         } else {
-            self.class_probabilities[class_index]
+            self.class_priors[class_index]
         }
     }
 
-    fn conditional_probability(&self, class_index: usize, j: &M::RowVector) -> T {
+    fn log_likelihood(&self, class_index: usize, j: &M::RowVector) -> T {
         if class_index < self.class_labels.len() {
-            let mut prob = T::one();
+            let mut likelihood = T::zero();
             for feature in 0..j.len() {
                 let value = j.get(feature);
                 match self.feature_categories[feature]
                     .iter()
                     .position(|&t| t == value)
                 {
-                    Some(_i) => prob *= self.coef[class_index][feature][_i],
+                    Some(_i) => likelihood += self.coef[class_index][feature][_i],
                     None => return T::zero(),
                 }
             }
-            prob
+            likelihood
         } else {
             T::zero()
         }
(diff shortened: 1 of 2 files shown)


VolodymyrOrlov:
It looks like something is wrong with the way your calculations. Could you please values compare values in `class_priors` and `coef` generated by this test:

```
    #[test]
    fn run_base_naive_bayes1() {
        let x = DenseMatrix::from_2d_array(&[
            &[3., 4., 0., 1.],
            &[3., 0., 0., 1.],
            &[4., 4., 1., 2.],
            &[4., 2., 4., 3.],
            &[4., 2., 4., 2.],
            &[4., 1., 1., 0.],
            &[1., 1., 1., 1.],
            &[0., 4., 1., 0.],
            &[0., 3., 2., 1.],
            &[0., 3., 1., 1.],
            &[3., 4., 0., 1.],
            &[3., 4., 2., 4.],
            &[0., 3., 1., 2.],
            &[0., 4., 1., 2.]
        ]);
        let y = vec![0., 0., 1., 1., 1., 0., 1., 0., 1., 1., 1., 1., 1., 0.];

        let cnb = CategoricalNB::fit(&x, &y, Default::default()).unwrap();
        println!("{:?}", cnb);
    }
```

with Scikit-learn:
image
for example, these is how values for class 1 and 0, category 0 and column 0 are calculated in Scikit-learn:
image

Your numbers are different

VolodymyrOrlov:
I've also noticed that neither `CategoricalNB` nor `CategoricalNBDistribution` implement `Serialize`, `Deserialize` and `Debug`. Every single algorithm in SmartCore implements these interfaces. Could you please make sure that these interfaces are implemented? Feel free to do it as a separate PR.

On top of it we need a test that verifies that your algorithms can be serialized and deserialized with Serde, similar to https://github.com/smartcorelib/smartcore/blob/development/src/linear/ridge_regression.rs#L295

morenol:
I fixed a bug that added when I was rewriting the code. 

Now y_hat for that test is the same that the y_hat that return scikit_learn.

 On the other hand I noticed that in the sciki_tlearn implementation the category_count is taking in consideration  that are not present in the data, i.e: category_count is always of size 5, because they are considering [0,1,2,3,4] for each category.  In the current implementation of smartcore we are only counting the values that are already present for each category. So for this particular data, instead of 

```python
>>> clf.category_count_
[array([[2., 0., 0., 2., 1.],
       [3., 1., 0., 2., 3.]]), array([[1., 1., 0., 0., 3.],
       [0., 1., 2., 3., 3.]]), array([[2., 3., 0., 0., 0.],
       [1., 4., 2., 0., 2.]]), array([[2., 2., 1., 0., 0.],
       [0., 4., 3., 1., 1.]])]
```
we have:
```rust
[[[2, 0, 2, 1], [3, 1, 2, 3]],
[[1, 1, 0, 0, 3],  [0, 1, 2, 3, 3]],
[[2, 3, 0, 0], [1, 4, 2, 2]],
[[2, 2, 1, 0, 0], [0, 4, 3, 1, 1]]]
```
For that reason the coefficients are different, i.e. 
As you said in the previous message, for scikit_learn in the class 0 and category 0 it is being calculated as

`ln ( (2 + 1) / ( 5 + 5))`

and currently in smartcore it is being calculated as:

`ln (( 2 + 1) / (5 + 4))`

What do you think @VolodymyrOrlov? Should I change the why we calculate the possible categories? It seems to me that scikit_learn is using the the max value of each feature to define which are the possible categories  (In fact I just checked that CategoricalNB does not accept features with negative values.)

VolodymyrOrlov:
I think that Scikit's implementation assumes that categories are encoded with ordinal numbers that's why there are 5 categories for every column of the data despite the fact that there are only samples for 4 categories in data for the first column: 0, 1, 3, 4. Because of this assumption Scikit learn thinks there is additional category (2) for which we have no example. 

The simplest way would be to use this assumption in SmartCore as well but the problem is that in SmartCore data is an array of floats and there is infinity many numbers between any pair of numbers. Hence you can't apply the same logic because you'll get infinite number of categories for every column. I see 2 solutions to this problem:
1. Cast `X` into a matrix of u64 (u32) or usize. This can be done once when you fit your algorithms to the data
2. Relax Scikit's assumption and instead assume that categories encoded as nominal values with each value being an ID. But in this case I would check the literature to better understand why Scikit learn assumes categories are ordinal numbers. Maybe there is some practical benefit in doing so.

morenol:
I like the first idea, it would allow us to use hashmaps and other data structures to make everything faster, and in fact I tried to run in scikit using 

```python
x = np.array([
            [3.4, 4.4, 0.4, 1.5],
            [3.2, 0.6, 0.7, 1.8],
            [4.5, 4.6, 1.5, 2.8],
            [4., 2.2, 4.5, 3.5],
            [4.7, 2., 4.1, 2.7],
            [4.2, 1.5, 1.7, 0.8],
            [1.3, 1.6, 1.7, 1.8],
            [0.4, 4., 1., 0.6],
            [0.7, 3.3, 2.4, 1.4],
            [0.4, 3.8, 1.6, 1.5],
            [3.2, 4.8, 0.9, 1.7],
            [3.1, 4.1, 2.7, 4.2],
            [0.6, 3.7, 1., 2.4],
            [0.4, 4.9, 1.4, 2.8]])
y = [0,0,1,1,1,0,1,0,1,1,1,1,1,0]
```
And I have the same results that you had, so I think that it is using the floor of each number. I will start to implement that.

VolodymyrOrlov (approved):
LGMT

PR #30, opened 2020-11-18
Add serde to CategoricalNB

morenol:


src/naive_bayes/categorical.rs
```diff
@@ -269,4 +269,32 @@ mod tests {
...
+            &[3., 4., 0., 1.],
+            &[3., 4., 2., 4.],
+            &[0., 3., 1., 2.],
+            &[0., 4., 1., 2.],
+        ]);
+
+        let y = vec![0., 0., 1., 1., 1., 0., 1., 0., 1., 1., 1., 1., 1., 0.];
+        let cnb = CategoricalNB::fit(&x, &y, Default::default()).unwrap();
+
+        let deserialized_cnb: CategoricalNB<f32, DenseMatrix<f32>> =
```
morenol, line 295: This works fine, but as long as I change this to f64 it fails due to floating floating precision

VolodymyrOrlov: That's because derived implementation of PartialEq compares floats directly. Instead, you should compare floats up to some precision, e.g. 
```
        let a:f64 = 10.0 + 1e-9;
        let b:f64 = 10.0 + 1e-9;
        let tolerance = std::f64::EPSILON;
        // comparing a and be using == would be wrong, because float is an approximation of a real number
        (a - b).abs() <= tolerance;
```
check out how I [implement PartialEq for RidgeRegression](https://github.com/smartcorelib/smartcore/blob/development/src/linear/ridge_regression.rs#L110)

src/naive_bayes/categorical.rs
```diff
@@ -6,14 +6,13 @@ use crate::naive_bayes::{BaseNaiveBayes, NBDistribution};
...
 
 /// Naive Bayes classifier for categorical features
-#[derive(Debug)]
-struct CategoricalNBDistribution<T: RealNumber> {
+#[derive(Serialize, Deserialize, Debug, PartialEq)]
+struct CategoricalNBDistribution<T: RealNumber, M: Matrix<T>> {
     class_labels: Vec<T>,
     class_priors: Vec<T>,
-    coefficients: Vec<Vec<Vec<T>>>,
+    coefficients: Vec<Vec<M>>,
```
VolodymyrOrlov, line 13: Can you use M::RowVector here instead? This will let you avoid code like this:
```
let coef_i_j = M::RowVector::from_array(&coef_i_j);
let coef_i_j = M::from_row_vector(coef_i_j);
```

VolodymyrOrlov (changes requested):

src/naive_bayes/categorical.rs
morenol: The problem with that is that RowVector does not have a PartialEq implementation, let me check if I can do that

VolodymyrOrlov: I think we should add `+ PartialEq` to BaseVector

morenol: Another associated problem is this:

```rust
13 |     coefficients: Vec<Vec<M::RowVector>>,
   |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ the trait `error::_::_serde::Deserialize<'_>` is not implemented for `<M as linalg::BaseMatrix<T>>::RowVector`
   |
```

morenol: I would have to implement a custom way to serialize and deserialize CategoricalNBDistribution, right?

VolodymyrOrlov: I've seen it before. [This solution](https://github.com/smartcorelib/smartcore/blob/development/src/svm/svc.rs#L106) seems to help

VolodymyrOrlov: On the other hand, you don't have to go through the trouble of storing coefficients as `Vec<Vec<M::RowVector>>`. You can keep using `Vec<Vec<Vec<T>>>` as you did before and define custom `PartialEq` for your `CategoricalNBDistribution` that has a method `eq`
```
impl<T: RealNumber> PartialEq for KMeans<T> {
    fn eq(&self, other: &Self) -> bool {
        ...
        for i in 0..self.coefficients.len() {
               for j in 0..self.coefficients[j] {
                    for k in 0..self.coefficients[j][k] {
                        if (self.coefficients[i][j][k] - other.centroids[i][j][k]).abs() > T::epsilon() {
                           return false;
                        }
                }
         ...
    }
}
```

morenol: I applied your suggestion and it worked fine but the tests are still failing due to the default PartialEq implementation of Vec<_>. I am going to follow your suggestion and use a custom implementation of PartialEq for CategoricalNBDistribution as I previously did in https://github.com/smartcorelib/smartcore/pull/30/commits/a23e66c5dc15ef0d604afbac6dee21f27f69feca

VolodymyrOrlov (approved):
lgtm

PR #33, opened 2020-11-23
Add benches for GNB

morenol:


benches/naive_bayes.rs
```diff
@@ -0,0 +1,73 @@
...
+                .collect::<Vec<f64>>();
+            group.bench_with_input(
+                BenchmarkId::from_parameter(format!(
+                    "n_samples: {}, n_features: {}",
+                    n_samples, n_features
+                )),
+                n_samples,
+                |b, _| {
+                    b.iter(|| {
+                        GaussianNB::fit(&x, &y, Default::default()).unwrap();
```
VolodymyrOrlov, line 28: I do not really understand value of [`black_box`](https://bheisler.github.io/criterion.rs/criterion/fn.black_box.html) but wondering, should we use it here for `x` and `y`. What do you think?

VolodymyrOrlov (dismissed):
Looks good! So what are the next steps? Were these tests helpful in highlighting bottlenecks and determining optimization plan?

VolodymyrOrlov (approved):

PR #34, opened 2020-11-23
Change implementation of to_row_vector for nalgebra

morenol:
Changing the implementation of to_row_vector in nalgebra. It is now a reshape to (1, number of elements).

It was necessary an upgrade of the nalgebra crate in order to be able to use `reshape_generic`.


VolodymyrOrlov (changes requested):
Can you also add a test that covers the corner case you've discovered?

VolodymyrOrlov (approved):

PR #35, opened 2020-11-25
LASSO

VolodymyrOrlov:
In this PR I add LASSO and some additional methods:

* Biconjugate gradient solver for solving a systems of linear equations
* Additional function `ab` for multiplying transposed matrices. 

Also, I fix small bug in matrix `dot` function

### DIFF
--- a/src/linalg/naive/dense_matrix.rs
+++ b/src/linalg/naive/dense_matrix.rs
@@ -9,6 +9,7 @@ use serde::{Deserialize, Serialize};
 
 use crate::linalg::cholesky::CholeskyDecomposableMatrix;
 use crate::linalg::evd::EVDDecomposableMatrix;
+use crate::linalg::high_order::HighOrderOperations;
 use crate::linalg::lu::LUDecomposableMatrix;
 use crate::linalg::qr::QRDecomposableMatrix;
 use crate::linalg::stats::MatrixStats;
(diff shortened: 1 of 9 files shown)


src/linalg/naive/dense_matrix.rs
```diff
@@ -444,6 +445,38 @@ impl<T: RealNumber> LUDecomposableMatrix<T> for DenseMatrix<T> {}
 
 impl<T: RealNumber> CholeskyDecomposableMatrix<T> for DenseMatrix<T> {}
 
+impl<T: RealNumber> HighOrderOperations<T> for DenseMatrix<T> {
+    fn ab(&self, a_transpose: bool, b: &Self, b_transpose: bool) -> Self {
```
morenol, line 449: Is there a specific reason to not use the default implementation of `fn ab`?

VolodymyrOrlov: the default implementation is less efficient because you run 2 for loops every transpose. In the worst case scenario, when you have to do 2 transposes you will end up doing 6 for loops. In addition to that every transpose creates a new 2d array. 
Bundling matrix operations together can improve performance since you can do many transformations in the same iteration. Take a look at https://nalgebra.org/quick_reference/#blas-operations

morenol (approved):

PR #39, opened 2020-12-16
fix: Update ndarray version

morenol:
Use the last version of ndarray, a change was necessary due to a change in the API:

> The old function stack has been renamed to concatenate. A new function stack with numpy-like semantics have taken its place. Old usages of stack should change to use concatenate.

See in: https://github.com/rust-ndarray/ndarray/blob/master/RELEASES.md


VolodymyrOrlov (approved):
👍

PR #42, opened 2020-12-18
Derive clone for NB Parameters structs,

morenol:


morenol:
This is a simple change to derive Clone for these structs, I needed that in https://github.com/morenol/python-smartcore/blob/65ac67f48701a97fe52f6b55c5b08989ae0648cf/src/naive_bayes.rs#L57

VolodymyrOrlov (approved):

Issue #58, opened 2021-01-05
Implement OneHotEncoder


VolodymyrOrlov:
Implement [OneHotEncoder](https://scikit-learn.org/stable/modules/generated/sklearn.preprocessing.OneHotEncoder.html#sklearn.preprocessing.OneHotEncoder), make logic similar to Scikit Learn's

The new encoder should belong to new module preprocessing and produce results on par with Scikit Learn

VolodymyrOrlov:
I think we can close this issue for now, thanks, @gaxler!

Issue #59, opened 2021-01-05
Implement OrdinalEncoder


VolodymyrOrlov:
Implement [OrdinalEncoder](https://scikit-learn.org/stable/modules/generated/sklearn.preprocessing.OrdinalEncoder.html#sklearn.preprocessing.OrdinalEncoder), make logic similar to Scikit Learn's

The new encoder should belong to new module preprocessing and produce results on par with Scikit Learn

gaxler:
As of now we limit categorical encoding to round RealNumbers, so for now OrdinalEncoder will be an identity function.

I'll adjust the SeriesOneHotEncoder so it will be able to make an ordinal encoding as well (with a more general CategoryMapper object). But won't implement any Ordinal transformers for now.

PR #76, opened 2021-01-26
One hot encoder

gaxler:
This PR implements OneHotEncoder for single class per label (#58 (comment))

Encoder supports any Hash+Clone types as labels and prduces RealNumber vectors.

Will wait for feedback before implementing multi-class encoders.


src/preprocessing/target_encoders.rs
```diff
@@ -0,0 +1,233 @@
...
+pub fn make_one_hot<T: RealNumber>(label_idx: usize, num_labels: usize) -> Vec<T> {
+    let (pos, neg) = (T::from_f64(1f64).unwrap(), T::from_f64(0f64).unwrap());
+    (0..num_labels)
+        .map(|idx| if idx == label_idx { pos } else { neg })
+        .collect()
+}
+
+impl<'a, LabelType: Hash + Eq + Clone> OneHotEncoder<LabelType> {
+    /// Fit an encoder to a lable list
+    pub fn fit(labels: &[LabelType]) -> Self {
```
VolodymyrOrlov, line 68: Can you please rename this method? Method `fit` is reserved for an operation that estimates parameters of a transformer from n-dimensional array.  Here is a good example of how method `fit` should look like https://github.com/smartcorelib/smartcore/blob/development/src/decomposition/pca.rs#L119

src/preprocessing/target_encoders.rs
```diff
@@ -0,0 +1,233 @@
...
+        Self::from_label_def(LabelDefinition::LabelToClsNumMap(labels))
+    }
+    /// Build an encoder from a predefined positional label-class num vector
+    pub fn from_positional_label_vec(labels: Vec<LabelType>) -> Self {
+        Self::from_label_def(LabelDefinition::PositionalLabel(labels))
+    }
+
+    /// Transform a slice of label types into one-hot vectors
+    /// None is returned if unknown label is encountered
+    pub fn transform<U: RealNumber>(&self, labels: &[LabelType]) -> Vec<Option<Vec<U>>> {
```
VolodymyrOrlov, line 98: Can you please rename this method? Method `transform ` is reserved for an operation that 2-dimensional array and transforms every single row using one-hot encoding.
Here is a good example of how method `transform` should look like https://github.com/smartcorelib/smartcore/blob/development/src/decomposition/pca.rs#L126

VolodymyrOrlov (changes requested):
Hi @gaxler, thank you for contributing your time to work on #58!

I think this PR is close to be merged into the development branch but I would like to request a couple of additional changes from you if you don't mind. 

I'd like to make sure that API of the `OneHotEncoder ` is in alignment with API of the rest of algorithms, as well as with API of Scikit learn. 

Take a look at the [`api`](https://github.com/smartcorelib/smartcore/blob/development/src/api.rs) module that summarizes interface and functions that are used to manipulate data and perform machine learning in Smartcore. Our API was lagely modeled after Scikit's API. 

Another good place to look is [`Scikit's OneHotEncoder`](https://scikit-learn.org/stable/modules/generated/sklearn.preprocessing.OneHotEncoder.html). 

What we want is to have methods `fit` and `transform` that can be used to and transform fit entire dataframe. 

This API compartability with Scikit makes easier switching to Smartcore as well as will let us integrate with Scikit Learn later down the road.

Feel free to add any additional functions that you feel are useful for Smartcore users

src/preprocessing/target_encoders.rs
```diff
@@ -0,0 +1,233 @@
...
+    num_classes: usize,
+}
+
+enum LabelDefinition<T> {
+    LabelToClsNumMap(HashMap<T, usize>),
+    PositionalLabel(Vec<T>),
+}
+
+/// Crearte a vector of size num_labels with zeros everywhere and 1 at label_idx (one-hot vector)
+pub fn make_one_hot<T: RealNumber>(label_idx: usize, num_labels: usize) -> Vec<T> {
```
morenol, line 59: I think that this should be generic, and return an implementation of smartcore::linalg::BaseVector;

gaxler:
@VolodymyrOrlov Thank You for taking the time to review this, appreciate the feed-back!

Let me make sure that I fully understand your comments:

We do want to have a consistent api with `fit` and `transform` for one-hot encoding, but those must act on entire dataframes as input.  My implementation acts on a series (if we stick to the dataframe analogy), so I'll rename those. and make the `fit` and `transform`  methods to act on a dataframe.

I'm a bit confused about what the dataframe should be in this case.  smartcore models act on vectors and matricies of `RealNumber` traits. I attempted to replicate Scikit's implementation where by and large you can encode any type of category.

The simplest solution I can think of is to make `OneHotEncoder` support only `RealNumber` categories and have the user define which ones are categorical.
What do you think?

src/preprocessing/target_encoders.rs
VolodymyrOrlov: > @VolodymyrOrlov Thank You for taking the time to review this, appreciate the feed-back!
> 
> Let me make sure that I fully understand your comments:
> 
> We do want to have a consistent api with `fit` and `transform` for one-hot encoding, but those must act on entire dataframes as input. My implementation acts on a series (if we stick to the dataframe analogy), so I'll rename those. and make the `fit` and `transform` methods to act on a dataframe.
> 
> I'm a bit confused about what the dataframe should be in this case. smartcore models act on vectors and matricies of `RealNumber` traits. I attempted to replicate Scikit's implementation where by and large you can encode any type of category.
> 
> The simplest solution I can think of is to make `OneHotEncoder` support only `RealNumber` categories and have the user define which ones are categorical.
> What do you think?

@gaxler This is correct, we'd like to support both, series and dataframes. 

I agree with you. `OneHotEncoder` should one-hot-encode not only floats but also other types, like integers and strings. The problem is that current design is not flexible enough to support all these types.

There is an easy and hard paths, depending on how much time you have to work on this feature. 

The easy path is to convert only [nominal and ordinal](https://towardsdatascience.com/data-types-in-statistics-347e152e8bee) floats. In this case you don't have to refactor Smartcore interfaces. You can either infer categorical columns from data (e.g. by taking a random sample and looking at values in the sample), or take a list of column IDs that should be encoded from user, as you've suggested. In fact, I think we should do both.

The hard path is to introduce two new types: dataframe and series. These types will be modeled after Pandas DataFrame and Series and will encapsulate data as a matrix (or as a vector) and additional metadata, like column names and types. Later we can switch all Smartcore algorithms to use these new types instead of matrices and vectors. Since most algorithms cannot work with categorical data directly we'll have to implement internal converters that will extract data from dataframe and series and transform these values into floats or throw an error, if it is not possible.

Let me know what you think about these options. Also, feel free to connect with me in Discord, if you have an account there. My ID is volodymyr.orlov#7062

VolodymyrOrlov:
> @VolodymyrOrlov Thank You for taking the time to review this, appreciate the feed-back!
> 
> Let me make sure that I fully understand your comments:
> 
> We do want to have a consistent api with `fit` and `transform` for one-hot encoding, but those must act on entire dataframes as input. My implementation acts on a series (if we stick to the dataframe analogy), so I'll rename those. and make the `fit` and `transform` methods to act on a dataframe.
> 
> I'm a bit confused about what the dataframe should be in this case. smartcore models act on vectors and matricies of `RealNumber` traits. I attempted to replicate Scikit's implementation where by and large you can encode any type of category.
> 
> The simplest solution I can think of is to make `OneHotEncoder` support only `RealNumber` categories and have the user define which ones are categorical.
> What do you think?

@gaxler This is correct, we'd like to support both, series and dataframes.

I agree with you. OneHotEncoder should one-hot-encode not only floats but also other types, like integers and strings. The problem is that current design is not flexible enough to support all these types.

There is an easy and hard paths, depending on how much time you have to work on this feature.

The easy path is to convert only [nominal and ordinal](https://towardsdatascience.com/data-types-in-statistics-347e152e8bee) floats. In this case you don't have to refactor Smartcore interfaces. You can either infer categorical columns from data (e.g. by taking a random sample and looking at values in the sample), or take a list of column IDs that should be encoded from user, as you've suggested. In fact, I think we should do both.

The hard path is to introduce two new types: dataframe and series. These types will be modeled after Pandas DataFrame and Series and will encapsulate data as a matrix (or as a vector) and additional metadata, like column names and types. Later we can switch all Smartcore algorithms to use these new types instead of matrices and vectors. Since most algorithms cannot work with categorical data directly we'll have to implement internal converters that will extract data from dataframe and series and transform these values into floats or throw an error, if it is not possible.

Let me know what you think about these options. Also, feel free to connect with me in Discord, if you have an account there. My ID is volodymyr.orlov#7062

gaxler:
I think I'll start with the easy path, just to make it usable. Afterwards, would love to work on the new data types.

src/preprocessing/target_encoders.rs
gaxler: Thanks for the comment
see: https://github.com/smartcorelib/smartcore/pull/76/commits/244a72444520cc6ac832779a44538fc93f6b68e3

VolodymyrOrlov (dismissed):
@gaxler LGTM! Let me know if you would like me to merge this PR into development or prefer to keep it open for a while.

gaxler:
Thanks!
It seems that we are really close to also doing https://github.com/smartcorelib/smartcore/issues/59 
Maybe we can do it in a single PR?

src/preprocessing/mod.rs
```diff
@@ -0,0 +1,5 @@
+/// Transform a data matrix by replaceing all categorical variables with their one-hot vector equivalents
+pub mod categorical_encoder;
```
VolodymyrOrlov, line 2: The module name is a bit too long. Can you rename it to either `categorical` or `one_hot_encoder`?

src/preprocessing/categorical_encoder.rs
```diff
@@ -0,0 +1,329 @@
...
+
+use crate::error::Failed;
+use crate::linalg::Matrix;
+
+use crate::preprocessing::data_traits::{CategoricalFloat, Categorizable};
+use crate::preprocessing::series_encoder::CategoryMapper;
+
+/// OneHotEncoder Parameters
+#[derive(Debug, Clone)]
+pub struct OneHotEncoderParams {
```
VolodymyrOrlov, line 37: Would you mind implementing a builder and `Default` for this struct? Something similar to https://github.com/smartcorelib/smartcore/blob/development/src/decomposition/pca.rs#L98?
Right now you have one parameter but with many parameters instantiation of the struct may be a bit easier to do if you have default + builder.

VolodymyrOrlov (dismissed):

src/preprocessing/categorical_encoder.rs
gaxler: This will be a bit of a problem. The parameter doesn't have any reasonable defaults. It just indicates what columns of a matrix represent categories

VolodymyrOrlov: I see. In this case please ignore this suggestion.

VolodymyrOrlov (approved):

PR #78, opened 2021-02-07
Implement Display for Dataset.

tormeh:
Rough first draft. Happy to get feedback


src/dataset/mod.rs
```diff
@@ -134,3 +134,62 @@ mod tests {
         assert_eq!(*m[1][3], 9);
     }
 }
+
+impl<X: Copy + std::fmt::Debug, Y: Copy + std::fmt::Debug> std::fmt::Display for Dataset<X, Y> {
```
VolodymyrOrlov, line 138: Can you move this block above the section with tests, please?

src/dataset/mod.rs
```diff
@@ -134,3 +134,62 @@ mod tests {
...
+        struct Feature<X> {
+            name: String,
+            value: X,
+        }
+        struct DataPoint<X, Y> {
+            labels: Vec<Target<Y>>,
+            features: Vec<Feature<X>>,
+        }
+        impl<X: Copy + std::fmt::Debug, Y: Copy + std::fmt::Debug> std::fmt::Display for DataPoint<X, Y> {
+            fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
```
VolodymyrOrlov, line 153: Can you add a simple unit test that covers this function? For example, I am getting an error when I try to print Iris Dataset:
```
  #[test]
  fn display() {
      let dataset = iris::load_dataset();

      println!("{}", dataset);
  }
````

File: src/dataset/mod.rs [hunk 3]
```diff
@@ -134,3 +134,62 @@ mod tests {
...
+        for sample_index in 0..self.num_samples {
+            let mut features = Vec::new();
+            for feature_index in 0..self.feature_names.len() {
+                features.push(Feature {
+                    name: self.feature_names[feature_index].to_owned(),
+                    value: self.data[sample_index * self.num_features + feature_index],
+                });
+            }
+            let mut targets = Vec::new();
+            for target_index in 0..self.target_names.len() {
```
VolodymyrOrlov, line 178: What if a dataset does not has names for all target variables? What if the `self.target_names` list does not cover all target variables?

src/dataset/mod.rs
```diff
@@ -134,3 +134,62 @@ mod tests {
...
+                        .iter()
+                        .map(|feature| format!("{}:{:?}, ", feature.name, feature.value))
+                        .collect::<String>()
+                )
+            }
+        }
+        let mut datapoints = Vec::new();
+        for sample_index in 0..self.num_samples {
+            let mut features = Vec::new();
+            for feature_index in 0..self.feature_names.len() {
```
VolodymyrOrlov, line 171: How do we handle datasets with empty `self.feature_names`?

src/dataset/mod.rs
```diff
@@ -134,3 +134,62 @@ mod tests {
...
+                write!(
+                    f,
+                    "{} : {}",
+                    self.labels
+                        .iter()
+                        .map(|target| format!("{}:{:?}, ", target.name, target.value))
+                        .collect::<String>(),
+                    self.features
+                        .iter()
+                        .map(|feature| format!("{}:{:?}, ", feature.name, feature.value))
```
VolodymyrOrlov, line 163: I am wondering, how will the proposed message format fits large datasets? Let's say we have 2K samples in a dataset. Do we need to repeat feature name for every value we display? It seems redundant to me...
Another thought, should we truncate the message when certain number of samples has been reached?

VolodymyrOrlov (changes requested):
Thanks, @tormeh! I agree we need a way to quickly display content of a dataset. And I like proposed approach in general, but I'd like visualization logic to cover these corner cases:

1. Attributes `self.target_names`  and `self.feature_names` are optional. We should be able to display datasets where both these attributes are set to an empty list
2. How can we compactly display large datasets?

src/dataset/mod.rs
tormeh: You could imagine a CSV-like format, But as someone who always chooses expanded output in Postgres, I'm not really a fan. It works really well until rows won't fit on a single line on the screen, and then it's awful. You could imagine that we intelligently choose depending on number of features vs. number of data points, which could be better, but not sure how it would work in practice. Maybe we should detect screen width vs num_features and decide based on that?

Truncating the message seems like a straightforward thing to do. Like, say, after 200 lines or something? I bet someone would want it configurable, though.

How do alternatives like Pandas (or whatever, I'm not familiar with data science tools) do it?

src/dataset/mod.rs
tormeh: Hm... I see the problem. I assumed (favourite word right there) that you would have a vector of targets per sample, and not use different integer values to represent different targets.
So, for a single sample example:
feature_vec * weight_vec = target_vec = [0, 1, 0, 0]
not:
feature_vec * weight_vec = target_int = 2

This worked in my case since I only had one target

src/dataset/mod.rs
tormeh: I've made a change that just uses debug print in that case

src/dataset/mod.rs
VolodymyrOrlov: Yes, `data` is a flattened matrix of size N x P, where N is number of samples and P is number of features per sample, `target` is a 1d array with target values of size P. I think printing raw values of `target` is a good solution.

src/dataset/mod.rs
VolodymyrOrlov: I see your point here, Tormod. It is always frustrating when you have more attributes per sample than your screen can fit and data shifts by one line making it harder to skim through the data. But I don't think that printing feature name along with every value makes it easier to read the data either. I've played with your code a little and made it print Iris dataset. This is what I see, please correct me it this is not how you intended the format to be in this PR: 

```
Iris dataset: https://archive.ics.uci.edu/ml/datasets/iris
setosa:0.0, versicolor:0.0, virginica:0.0, :0.0,  : sepal length (cm):5.1, sepal width (cm):3.5, petal length (cm):1.4, petal width (cm):0.2, 
setosa:0.0, versicolor:0.0, virginica:0.0, :0.0,  : sepal length (cm):4.9, sepal width (cm):3.0, petal length (cm):1.4, petal width (cm):0.2, 
setosa:0.0, versicolor:0.0, virginica:0.0, :0.0,  : sepal length (cm):4.7, sepal width (cm):3.2, petal length (cm):1.3, petal width (cm):0.2, 
setosa:0.0, versicolor:0.0, virginica:0.0, :0.0,  : sepal length (cm):4.6, sepal width (cm):3.1, petal length (cm):1.5, petal width (cm):0.2, 
setosa:0.0, versicolor:0.0, virginica:0.0, :0.0,  : sepal length (cm):5.0, sepal width (cm):3.6, petal length (cm):1.4, petal width (cm):0.2, 
```

One problem with this way of displaying data is that it is hard to focus on numbers due to repeated feature names. Also, if we have more attributes per sample data will still take more lines per sample thus making it harder to see where one sample ends and another starts.

Pandas displays data as a table, that is truncated to `max_rows` and `max_columns`:

image

PR #79, opened 2021-02-15
fix: Fix new clippy warnings

morenol:
This fixes the clippy warnings added in the new version of clippy

### DIFF
--- a/src/linalg/mod.rs
+++ b/src/linalg/mod.rs
@@ -1,3 +1,4 @@
+#![allow(clippy::wrong_self_convention)]
 //! # Linear Algebra and Matrix Decomposition
 //!
 //! Most machine learning algorithms in SmartCore depend on linear algebra and matrix decomposition methods from this module.
@@ -265,7 +266,7 @@ pub trait BaseVector<T: RealNumber>: Clone + Debug {
             sum += xi * xi;
         }
         mu /= div;
-        sum / div - mu * mu
+        sum / div - mu.powi(2)
     }
     /// Computes the standard deviation.
     fn std(&self) -> T {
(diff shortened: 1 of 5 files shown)


morenol:
There is still an error with this clippy::suspicious-operation-groupings. Could you check that @VolodymyrOrlov? It looks to me that we can ignore that warning

src/linalg/mod.rs
```diff
@@ -265,7 +266,7 @@ pub trait BaseVector<T: RealNumber>: Clone + Debug {
             sum += xi * xi;
         }
         mu /= div;
-        sum / div - mu * mu
+        sum / div - mu.powi(2)
```
VolodymyrOrlov, line 269: I am wondering what is faster in Rust, mu * mu or mu.powi(2)? I always thought that multiplication is faster but that might not be true in Rust. This might be a small thing, but if repeated many times the difference will accumulate to a significant value.

VolodymyrOrlov (dismissed):
lgtm

src/linalg/mod.rs
VolodymyrOrlov: nm, I found my answer here: https://stackoverflow.com/questions/38799724/is-there-a-performance-difference-between-a-squared-function-and-directly-callin

VolodymyrOrlov:
> There is still an error with this clippy::suspicious-operation-groupings. Could you check that @VolodymyrOrlov? It looks to me that we can ignore that warning

Agree, the warnings I see do not make any sense. The logic that is highlighted by the lint is correct

morenol:
Merging because it already was approved

Issue #53, opened 2021-01-05
Add new `make_moons` data generator


VolodymyrOrlov:
`make_moons` data generator will be useful to compare different clustering algorithms.  Implementation details and parameters can be found in [Scikit Learn](https://scikit-learn.org/stable/modules/generated/sklearn.datasets.make_moons.html#sklearn.datasets.make_moons). We need an exact copy of this function

morenol:
Added in #82

PR #81, opened 2021-02-18
Adding make_moons generators

cmccomb:
Implemented a make_moons generator in response to #53. Followed the same logic as [sklearn](https://github.com/scikit-learn/scikit-learn/blob/95119c13a/sklearn/datasets/_samples_generator.py#L683) and borrowed some syntax from the make_circles generator in smartcore.


cmccomb:
Fixing formatting

Issue #87, opened 2021-03-05
mod.rs deserialize_data uses architecture specific usize which errors on 32bit linux/windows


ronniec95:
```
pub(crate) fn deserialize_data(
    bytes: &[u8],
) -> Result<(Vec<f32>, Vec<f32>, usize, usize), io::Error> {
    // read the same file back into a Vec of bytes
    let (num_samples, num_features) = {
        let mut buffer = [0u8; 8];
        buffer.copy_from_slice(&bytes[0..8]);
        let num_features = usize::from_le_bytes(buffer); // This line does not compile on 32bit systems. Change to u64
        buffer.copy_from_slice(&bytes[8..16]);
        let num_samples = usize::from_le_bytes(buffer);
        (num_samples, num_features)
    };
  ...
}
```

morenol:
Thanks for the report @ronniec95, in case that you don't need the datasets feature, I think that you can avoid this error if you install smartcore without the 'datasets' feature which is in the defaults.

In https://github.com/smartcorelib/smartcore/pull/80 I already fixed a problem like this for wasm, but it looks like we have to create a more generic way to avoid the problem. I will look for better ways to solve this.

ronniec95:
Yeah I fixed it locally, just reporting for reference

ronniec95:
It appears to work, but you can see immediately that if the data was >2^32-1 in size it would wrap around.
For this example data, it's fine on win32 at least.

PR #88, opened 2021-03-05
fix: Use usize size for usize::from_le_bytes buffer

morenol:
Fixes https://github.com/smartcorelib/smartcore/issues/87


morenol:
I would appreciate @ronniec95 if you confirm me that this fixes the problem. I am going to check which images we can use to run the build tests with other architectures

morenol:
Before merging this @VolodymyrOrlov I would like to add a way in the CI to check this tool for different architectures and OS, what do you think if we move to Github Actions?

VolodymyrOrlov:
@morenol I am fine as long as Github Actions supports all build steps.

VolodymyrOrlov (approved):

Issue #85, opened 2021-02-26
Refactor linalg module


VolodymyrOrlov:
I want to use this issue to share a heads-up on a big refactoring that I plan for the `linalg` module. 

During last couple of month I've seen on multiple occasions limitations and shortcomings imposed by the current design of the `BaseVector` and `BaseMatrix`. To mention a couple here:

* It is not possible to define an instance of `BaseMatrix` that holds string, integer type values. 
* `BaseMatrix` is not designed to hold values that belong to multiple types
* Some algorithms, e.g. RandomForest, does not use most methods defined in the `BaseMatrix` and `BaseVector`. Some preprocessing methods that we plan for future, like LabelEncoder will not need linear algebra routines defined for both classes. 
* Some basic operations, like get row or get column, perform unnecessary copy. This problem stems from the fact that both structs do not provide views or iterators that lets developer access an internal structure of the data. 
* All operations are defined as functions. While this is not a big deal it leads to a clumsy looking code. Instead it would be nice to use more traits defined in [std::ops](https://doc.rust-lang.org/std/ops/index.html)

As a result, I'd like to see how can we use Rust's type system to design a better container for data that solves all these shortcomings. 

I am open to any suggestions you have. Feel free to post your ideas here.

gaxler:
As for the multi-type,
Perhaps we can use something like Apache Arraow's [arrays](https://docs.rs/arrow/3.0.0/arrow/array/index.html) to construct a multi-type dataframe  (https://docs.rs/arrow/3.0.0/arrow/record_batch/struct.RecordBatch.html)

VolodymyrOrlov:
@gaxler I like the idea of building 2- and 1- dimensional array on top of Apache Arrow, but I'd like it to be optional. The `linalg` module provides an abstraction layer around data, it should not force developers to use any concrete libraries like ndarray, nalgebra or apache arrow to hold the data. I'd like to keep it this way to make sure we can add new array types as easy as possible. 

Btw, last time  I've checked Arrow's Arrays are immutable. Not sure how we will use immutable array in algorithms that need to change a small portion of the data while leaving every other cell intact.

gaxler:
@VolodymyrOrlov I agree on Arrow being optional, I meant to use Arrow as some form of reference for multi-datatpye structures. 

`BaseDataFrame` trait will be the general API.

[...]

PR #108, opened 2021-05-31
First draft of the new n-dimensional arrays + NB use case

VolodymyrOrlov:
This PR will eventually solve #85. The goal is to have an abstract layer that can be implemented by concrete library that enables all traditional functionality provided by n-dimensional arrays: array manipulation, transformation, basic linear algebra routines, views e.t.c. 

The abstract layer I am proposing here enables us to:
* have arrays that can hold multiple types, including strings
* implement more efficient ML algorithms that minimize memory copy operations
* have ML algorithms that declare more granular requirements to input and output types, e.g. we can declare an input type for an algorithm that does need a simple non-mutable array w/o typical linear algebra routines. 

As a bonus, this PR introduces a row-major dense matrix that might get handy in some cases.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -66,9 +66,10 @@
 //! let y_hat = knn.predict(&x).unwrap();
 //! ```
 
+pub mod num;
+pub mod api;
 /// Various algorithms and helper methods that are used elsewhere in SmartCore
 pub mod algorithm;
-pub mod api;
 /// Algorithms for clustering of unlabeled data
 pub mod cluster;
 /// Various datasets
@@ -99,3 +100,5 @@ pub mod preprocessing;
 pub mod svm;
 /// Supervised tree-based learning methods
 pub mod tree;
+#[cfg(test)]
+pub mod utils;
--- a/src/naive_bayes/categorical.rs
+++ b/src/naive_bayes/categorical.rs
@@ -6,35 +6,36 @@
 //! Example:
 //!
 //! ```
-//! use smartcore::linalg::naive::dense_matrix::*;
+//! use smartcore::linalg::dense::matrix::DenseMatrix;
 //! use smartcore::naive_bayes::categorical::CategoricalNB;
 //!
 //! let x = DenseMatrix::from_2d_array(&[
-//!              &[3., 4., 0., 1.],
-//!              &[3., 0., 0., 1.],
-//!              &[4., 4., 1., 2.],
-//!              &[4., 2., 4., 3.],
-//!              &[4., 2., 4., 2.],
-//!              &[4., 1., 1., 0.],
-//!              &[1., 1., 1., 1.],
-//!              &[0., 4., 1., 0.],
-//!              &[0., 3., 2., 1.],
-//!              &[0., 3., 1., 1.],
-//!              &[3., 4., 0., 1.],
-//!              &[3., 4., 2., 4.],
-//!              &[0., 3., 1., 2.],
-//!              &[0., 4., 1., 2.],
+//!              &[3, 4, 0, 1],
+//!              &[3, 0, 0, 1],
+//!              &[4, 4, 1, 2],
+//!              &[4, 2, 4, 3],
+//!              &[4, 2, 4, 2],
+//!              &[4, 1, 1, 0],
+//!              &[1, 1, 1, 1],
+//!              &[0, 4, 1, 0],
+//!              &[0, 3, 2, 1],
+//!              &[0, 3, 1, 1],
+//!              &[3, 4, 0, 1],
+//!              &[3, 4, 2, 4],
+//!              &[0, 3, 1, 2],
+//!              &[0, 4, 1, 2],
 //!          ]);
-//! let y = vec![0., 0., 1., 1., 1., 0., 1., 0., 1., 1., 1., 1., 1., 0.];
+//! let y: Vec<u32> = vec![0, 0, 1, 1, 1, 0, 1, 0, 1, 1, 1, 1, 1, 0];
 //!
 //! let nb = CategoricalNB::fit(&x, &y, Default::default()).unwrap();
 //! let y_hat = nb.predict(&x).unwrap();
 //! ```
+use num_traits::Unsigned;
+
 use crate::api::{Predictor, SupervisedEstimator};
 use crate::error::Failed;
-use crate::linalg::BaseVector;
-use crate::linalg::Matrix;
-use crate::math::num::RealNumber;
+use crate::linalg::base::{Array2, Array1, ArrayView1};
+use crate::num::Number;
 use crate::naive_bayes::{BaseNaiveBayes, NBDistribution};
 #[cfg(feature = "serde")]
 use serde::{Deserialize, Serialize};
@@ -42,14 +43,14 @@ use serde::{Deserialize, Serialize};
 /// Naive Bayes classifier for categorical features
 #[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
 #[derive(Debug)]
-struct CategoricalNBDistribution<T: RealNumber> {
+struct CategoricalNBDistribution<T: Number + Unsigned> {
     /// number of training samples observed in each class
     class_count: Vec<usize>,
     /// class labels known to the classifier
     class_labels: Vec<T>,
     /// probability of each class
-    class_priors: Vec<T>,
-    coefficients: Vec<Vec<Vec<T>>>,
+    class_priors: Vec<f64>,
+    coefficients: Vec<Vec<Vec<f64>>>,
     /// Number of features of each sample
     n_features: usize,
     /// Number of categories for each feature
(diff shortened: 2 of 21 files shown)


VolodymyrOrlov:
@morenol there are multiple problems with this PR, like `println`s, lack of documentation, e.t.c. and I plan to fix these within next 3-4 of weeks. But I thought I'd share this PR with you so that you can take a look on proposed changes earlier. Besides, I've decided to try new abstract layer on NB first.

morenol (commented):
@VolodymyrOrlov this looks nice.

Since we are rethinking the way that the arrays are represented I think that we should consider the usage of  [const generics](https://www.infoq.com/news/2021/03/rust-1-51-released/).

src/naive_bayes/categorical.rs
```diff
@@ -6,50 +6,51 @@
...
+//!              &[0, 4, 1, 0],
+//!              &[0, 3, 2, 1],
+//!              &[0, 3, 1, 1],
+//!              &[3, 4, 0, 1],
+//!              &[3, 4, 2, 4],
+//!              &[0, 3, 1, 2],
+//!              &[0, 4, 1, 2],
 //!          ]);
-//! let y = vec![0., 0., 1., 1., 1., 0., 1., 0., 1., 1., 1., 1., 1., 0.];
+//! let y: Vec<u32> = vec![0, 0, 1, 1, 1, 0, 1, 0, 1, 1, 1, 1, 1, 0];
```
morenol, line 28: It is nice to be able to use normal integers for this

VolodymyrOrlov: The problem is that you make a call to `.to_usize()` in `fn log_likelihood` and `fn fit`, and this function will fail if used on negative integer values. Why would you like to support integers?

morenol: I am sorry, I was trying to say that it is good to be able to use other types different than floats (the constraint that we currently have in the main branch)

VolodymyrOrlov: I see! I thought you want to make sure we can run CategoricalNB on integer target variables. Yes, that's the plan! Floats are great, but we do need to support other types as well. That's one reason why I decided to try new layer on Naive Bayes first, since NB methods do not apply to floating-point data most of the time.

VolodymyrOrlov:
> @VolodymyrOrlov this looks nice.
> 
> Since we are rethinking the way that the arrays are represented I think that we should consider the usage of [const generics](https://www.infoq.com/news/2021/03/rust-1-51-released/).

Any concrete ideas how we can use const generics to improve the proposed abstract layer? For example, const generics can be used to define number of [dimensions in S](https://github.com/smartcorelib/smartcore/blob/2c7fab0a37c5acbce3d3fe154571bfa83f9890da/src/linalg/base.rs#L11)

morenol:
I think that I edited your comment instead of a response: here it is 

> Any concrete ideas how we can use const generics to improve the proposed abstract layer? For example, const generics can be used to define number of [dimensions in S](https://github.com/smartcorelib/smartcore/blob/2c7fab0a37c5acbce3d3fe154571bfa83f9890da/src/linalg/base.rs#L11)

I am not sure how we can use them in the abstract layer. I was thinking  in the structs defined in linalg::dense::. We can create an special type for static sized arrays. But maybe it is better to keep it simple and use nalgebra and ndarray when we want that kind of optimization

Mec-iS:
Great work Volodymyr! This looks really promising.
Just few notes to check if I understood the new abstraction and to help devs in digging:
* we are going to have new traits groups to define types for array/vectors elements: `Num`/`Number` and `FloatNumber` (I like it!) directly using `num_traits`;
* A lot of things going on in `linalg/base.rs`:
    * new smartcore `Array`-related type declarations to provide a unified interface, including the ones from optimised strides read/write. 
    * this interface is the toolbox to be leveraged by higher-level types, in particular the new implementations in `linalg/dense/`;
* we have now "views" types that, I assume, are to provide segmentation for the unerlying structs so to allow reusing of the same instantiated struct and avoid copying;
* supercool the "Decomposable"s for structs decomposition. 

Will help with documentation as soon as it is relatively stable.

@steboss check it out!

VolodymyrOrlov:
> Great work Volodymyr! This looks really promising.
> Just few notes to check if I understood the new abstraction and to help devs in digging:
> 
> * we are going to have new traits groups to define types for array/vectors elements: `Num`/`Number` and `FloatNumber` (I like it!) directly using `num_traits`;
> * A lot of things going on in `linalg/base.rs`:
>   
>   * new smartcore `Array`-related type declarations to provide a unified interface, including the ones from optimised strides read/write.
>   * this interface is the toolbox to be leveraged by higher-level types, in particular the new implementations in `linalg/dense/`;
> * we have now "views" types that, I assume, are to provide segmentation for the unerlying structs so to allow reusing of the same instantiated struct and avoid copying;
> * supercool the "Decomposable"s for structs decomposition.
> 
> Will help with documentation as soon as it is relatively stable.
> 
> @Steboss check it out!

That's right! The goal is to build a rich interface for 1- and 2-dimensional arrays that can be extended by anyone later. On one hand we can make ML algorithms more efficient by relying on a concrete library, like nalgebra or ndarray. Investing time into an abstract layer that facilitates integration with 3rd partly libraries, on the other hand, gives us a lot of flexibility and saves time later, if we find it necessary to bring in some other library in the future.

Mec-iS:
@morenol please take a look, there are still missing implementation but the code is at least cleaned up.

Mec-iS:
* conflicts with `development` should be now solved
* there are some types that are not correct, in particular in `src/linear/ridge_regression.rs` and `src/linear/logistic_regression.rs`
* tests need to checked against the new typed structures and implementations.
* ... (probably hundreds of things more) 💯

src/algorithm/neighbour/cover_tree.rs
```diff
@@ -60,36 +59,43 @@ impl<T, F: RealNumber, D: Distance<T, F>> PartialEq for CoverTree<T, F, D> {
 
 #[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
 #[derive(Debug)]
-struct Node<F: RealNumber> {
+struct Node {
     idx: usize,
+<<<<<<< HEAD
```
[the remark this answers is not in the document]
Mec-iS: I missed 1/75 ... ~0.1 percent, acceptable accuracy :D

morenol:
> please fetch this branch and run `cargo build --lib`, maybe there are some errors for which the solution is evident to you people.

I can take a look later today

src/lib.rs
```diff
@@ -7,6 +7,7 @@
 )]
 #![warn(missing_docs)]
 #![warn(rustdoc::missing_doc_code_examples)]
+#![feature(associated_type_bounds)]
```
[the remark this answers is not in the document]
Mec-iS: right. try to remove that line and see if you can fix the errors that pop up using the stable toolchain.

Mec-iS:
sure, when you have time. If we can fix the static analyzer (in `src/ensemble`) and compiler errors maybe we can move into running test 👍🏼

Mec-iS:
@montanalow please take a look to `ensemble` and `linear` modules as you are more familiar with those procedures. The objective is to make things to compile and then move into making the tests to pass. Also I saw some conflicts when pulling the very latest changes.

Mec-iS:
great job @morenol 
Can you tell me why we are missing definitions for `ElasticNetSearchParameters`, `LassoSearchParameters`, `LinearRegressionSearchParameters`? that instead are present in `development`?
I will try a merge, another hundred conflicts to solve

morenol:
> ElasticNetSearchParameters

not completely sure, we probably need to split in several PRs once we have all the conflicts solved in order to make it easy to review

Mec-iS:
@montanalow great! I was splitting the changes into multiple PRs but if you can do in one pass that would be awesome.

montanalow:
> @montanalow great! I was splitting the changes into multiple branches but if you can do in one pass that would be awesome.

#171 is up that should resolve all merge conflicts in this branch, although it's a fairly large change set...

Mec-iS:
The library now compiles but there are still a lot of runtime errors and tests to fix.

I have not yet a clear understanding about what it is going to be the substitute for `DenseMatrix`, I thought it should have been `Array2` but how to instantiate it? For example [this test](https://github.com/smartcorelib/smartcore/pull/108/files#diff-4987eecc531093812f2318ce76f233297739ac1fd034a6ca82f62ac5ec1b27d6R648). Also `predict_oob` is still broken as I don't know how to implement it without using the `samples` property that has been removed.

I am not still sure we can merge this all at once, I will probably start opening a PR to merge `num.rs` so we can do it gradually.

Mec-iS:
#173

Also I still don't understand how traits like `QRDecomposable` and the others are supposed to be used.

Mec-iS:
I am seeing discrepancies between the `metrics::recall` module in development and the one in this branch. Things are getting too entangled, please move and follow up to #173 for a more stepwise implementation.

Mec-iS:
please help. it is quite easy, just follow the implementations in the other ported modules. cc: @morenol @montanalow @titoeb 

Missing modules, 
* `preprocessing`
* `readers`
* `svm`

Mec-iS:
All work moved to https://github.com/smartcorelib/smartcore/tree/v0.5-wip

PR #120, opened 2021-11-10
 Allow setting seed for `RandomForestClassifier` and `Regressor`

### LINKED ISSUE #118: Allow setting seed for `RandomForestClassifier` and `Regressor`
To make them reproducible. This would include passing a RNG to `RandomForestClassifier::<T>::sample_with_replacement`:
https://github.com/smartcorelib/smartcore/blob/521dab49ef92a955eb860ca195c89feab17c4b59/src/ensemble/random_forest_classifier.rs#L223-L224
to be used instead of `rng::thread_rng()`:
https://github.com/smartcorelib/smartcore/blob/521dab49ef92a955eb860ca195c89feab17c4b59/src/ensemble/random_forest_classifier.rs#L307-L308
to subsample rows for each tree.

The same RNG would also need to be passed to `DecisionTreeClassifier::fit_weak_learner`:
https://github.com/smartcorelib/smartcore/blob/521dab49ef92a955eb860ca195c89feab17c4b59/src/ensemble/random_forest_classifier.rs#L235
and then to
https://github.com/smartcorelib/smartcore/blob/521dab49ef92a955eb860ca195c89feab17c4b59/src/tree/decision_tree_classifier.rs#L387
to be used in this `shuffle`:
https://github.com/smartcorelib/smartcore/blob/521dab49ef92a955eb860ca195c89feab17c4b59/src/tree/decision_tree_classifier.rs#L484-L486

Same for the `RandomForestRegressor`.

I can set something up and open a PR.

mlondschien:
Closes #118.

### DIFF
--- a/src/ensemble/random_forest_classifier.rs
+++ b/src/ensemble/random_forest_classifier.rs
@@ -45,10 +45,11 @@
 //!
 //! <script src="https://polyfill.io/v3/polyfill.min.js?features=es6"></script>
 //! <script id="MathJax-script" async src="https://cdn.jsdelivr.net/npm/mathjax@3/es5/tex-mml-chtml.js"></script>
+use rand::rngs::StdRng;
+use rand::{Rng, SeedableRng};
 use std::default::Default;
 use std::fmt::Debug;
 
-use rand::Rng;
 #[cfg(feature = "serde")]
 use serde::{Deserialize, Serialize};
 
@@ -79,6 +80,8 @@ pub struct RandomForestClassifierParameters {
     pub m: Option<usize>,
     /// Whether to keep samples used for tree generation. This is required for OOB prediction.
     pub keep_samples: bool,
+    /// Seed used for bootstrap sampling and feature selection for each tree.Failed
+    pub seed: u64,
 }
 
 /// Random Forest Classifier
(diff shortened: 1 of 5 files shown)


morenol (approved):

src/ensemble/random_forest_classifier.rs
```diff
@@ -79,6 +80,8 @@ pub struct RandomForestClassifierParameters {
     pub m: Option<usize>,
     /// Whether to keep samples used for tree generation. This is required for OOB prediction.
     pub keep_samples: bool,
+    /// Seed used for bootstrap sampling and feature selection for each tree.Failed
```
morenol, line 83: I think that we should remove `.Failed` here at the end

src/ensemble/random_forest_regressor.rs
```diff
@@ -75,6 +76,8 @@ pub struct RandomForestRegressorParameters {
     pub m: Option<usize>,
     /// Whether to keep samples used for tree generation. This is required for OOB prediction.
     pub keep_samples: bool,
+    /// Seed used for bootstrap sampling and feature selection for each tree.Failed
```
morenol, line 79: I think that we should remove `.Failed` here at the end

src/ensemble/random_forest_classifier.rs
mlondschien: Definitely! Not sure how I missed this. Thanks for the catch.

morenol (approved):

PR #137, opened 2022-07-09
Random forest parallelization

daraghk:


VolodymyrOrlov:
@daraghk , thank you very much for your contribution to Smartcore! The code looks good, the only thing that is left are `clippy` warnings and better test coverage. Once you have closed all pending checks we can merge this PR to `development`

daraghk:
@VolodymyrOrlov happy to contribute! I've fixed the code coverage issue but the linting errors are unrelated to any code I touched - you can see the same issues in the other random forest pr which is open at  the moment.

src/ensemble/random_forest_classifier.rs
```diff
@@ -221,41 +240,101 @@ impl<T: RealNumber> RandomForestClassifier<T> {
...
-                criterion: parameters.criterion.clone(),
-                max_depth: parameters.max_depth,
-                min_samples_leaf: parameters.min_samples_leaf,
-                min_samples_split: parameters.min_samples_split,
-            };
-            let tree =
-                DecisionTreeClassifier::fit_weak_learner(x, y, samples, mtry, params, &mut rng)?;
-            trees.push(tree);
-        }
+        let thread_pool = rayon::ThreadPoolBuilder::new()
```
morenol, line 246: It looks like this fail for WASM targets, Could you explore using this https://github.com/GoogleChromeLabs/wasm-bindgen-rayon? or keep the old behavior for `fit` when using `wasm` target, and this one of non `wasm` targets?.

morenol (changes requested):
Code looks good to me.

Just an issue with WASM targets and multithreading support.

Regarding the clippy issue, I think that a rebase/merge with `development` will fix that.

morenol:
I think that we could do parallelization with a feature flag that is disabled for WASM targets

Mec-iS:
I know but we have no bandwidth to follow up on all these features, creating a feature for one case can be only an increase in complexity with very few benefits. We need to prioritise or allocate more time or have somebody that carries on the task of extending parallelization to more algorithms.

Issue #50, opened 2021-01-04
Implement a new method that predicts probabilities, where it makes sense


VolodymyrOrlov:
Classification algorithms usually offer a way to quantify certainty of a prediction. In Scikit learn a method that returns probability estimates for all classes is called `predict_proba`. 

We need a similar method in SmartCore. One way to do it is to define a new trait `Classifier`, that will have a function `predict_proba`, and implement this trait for every algorithm where  predicts probabilities makes sense.

PR #138, opened 2022-07-11
Added per-class probability prediction for random forests

AlanRace:
Added a function to predict the per-class probability of each class for each observation.

```rust
let probabilities = forest.predict_probs(&data).unwrap();
```

`probabilities` is a _KxC_ matrix, where _K_ is the number of observations and _C_ is the number of classes. Probabilities are calculated as the fraction of trees in the random forest that predicted the given class.

Answer to #50 for random forests.


src/ensemble/random_forest_classifier.rs
```diff
@@ -316,6 +317,36 @@ impl<T: RealNumber> RandomForestClassifier<T> {
         which_max(&result)
     }
 
+    /// Predict the per-class probabilties for each observation. The probability is calculated as the fraction of trees that predicted a given class
+    pub fn predict_probs<M: Matrix<T>>(&self, x: &M) -> Result<DenseMatrix<f64>, Failed> {
```
vloplus, line 321: Could you please cover this and the next method with unit test? This will also help to get through automatic CI/CD checks

VolodymyrOrlov, line 321: Could you please cover this method, as well as the function `predict_probs_for_row` with unit test? This will help to get your code through automatic checks

VolodymyrOrlov:
@AlanRace thank you for your contribution to Smartcore! The change looks good, but you might want to look at `clippy` warnings as well as increase test coverage to get this code through automatic checks

Mec-iS:
https://github.com/AlanRace/smartcore/pull/1

AlanRace:
@Mec-iS Thanks for supplying the test - I am guessing there was a problem due to row-major vs column-major storage of `DenseMatrix`? Swapping the number of rows and columns in your test and then transposing the matrix results in a passing test.

Mec-iS:
thanks @AlanRace 

it is probably better to adhere to the DenseMatrix format, so it would be nice for the method to return the transposed values or directly a DenseMatrix.

AlanRace:
Maybe I am misunderstanding, but `predict_probs` does return a `DenseMatrix`.  

It looks like `DenseMatrix::from_vec` (which is called from `DenseMatrix::from_array` as part of your test) assumes that the given vector is in row-major form, but the entered values in the test are in column-major form. 

Would you prefer that the matrix returned from `predict_probs` is `num classes x num observations`, rather than the current `num observations x num classes`?

Mec-iS:
yeah, probably in the shape returned by `from_vector` and `from_array` is handier. thanks again

Mec-iS:
@morenol could you please take a look to the WASM test failing? it looks like we have different results for different targets.

src/ensemble/random_forest_classifier.rs
```diff
@@ -553,6 +554,37 @@ impl<T: RealNumber> RandomForestClassifier<T> {
         which_max(&result)
     }
 
+    /// Predict the per-class probabilties for each observation.
+    /// The probability is calculated as the fraction of trees that predicted a given class
+    pub fn predict_probs<M: Matrix<T>>(&self, x: &M) -> Result<DenseMatrix<f64>, Failed> {
```
morenol, line 559: in scikit it is called predict_proba, I think that it is better to keep the same name

src/ensemble/random_forest_classifier.rs
```diff
@@ -742,4 +774,70 @@ mod tests {
...
+        println!("{:?}", results);
+        println!("{:?}", results.shape());
+
+        assert_eq!(
+            results,
+            DenseMatrix::<f64>::from_array(
+                20,
+                2,
+                &[
+                    1.0, 0.0, 0.78, 0.22, 0.95, 0.05, 0.82, 0.18, 1.0, 0.0, 0.92, 0.08, 0.99, 0.01,
```
morenol, line 836: Are these the expected values?

Mec-iS: yes. those are the results as returned by the test. they match among all the targets except Wasm.

morenol: They are failing for me locally, and they also failed in the  CI for x86_64-unknown-linux-gnu, 

I think that the green checks in the CI are only in the CI jobs that builds the crate but not run tests

alexis2804:
Hello guys, when does this fucntion will be available ? I am totally need it in order to perform model ensembling !

Thanks a lot

Mec-iS:
moved to #138 to solve conflicts

PR #139, opened 2022-07-14
Fix clippy warnings for Rust 1.62.0

morenol:


morenol:
Merging because this is a two lines change PR without any impact for library callers

Issue #54, opened 2021-01-05
Implement StandardScaler


VolodymyrOrlov:
StandardScaler standardizes features by removing the mean and scaling to unit variance. Implementation details and parameters can be found in [Scikit Learn](https://scikit-learn.org/stable/modules/generated/sklearn.preprocessing.StandardScaler.html). 

The algorithm should be implemented as a struct that extends [`Transformer`](https://github.com/smartcorelib/smartcore/blob/development/src/api.rs#L39). The struct should belong to a new module `preprocessing`

Mec-iS:
I have done some work with scalers recently. So I can pick this up.

titoeb:
@Mec-iS: I had a look at this issue in my vacation and it resulted in this PR #143, feel free to comment and bring in the ideas you had in mind.

PR #143, opened 2022-08-24
Implementation of Standard scaler

titoeb:
In this branch I added an implementation for a `StandardScaler` as described in issue #54.

### DIFF
--- a/src/preprocessing/numerical.rs
+++ b/src/preprocessing/numerical.rs
@@ -1,0 +1,404 @@
+//! # Standard-Scaling For [RealNumber](../../math/num/trait.RealNumber.html) Matricies
+//! Transform a data [Matrix](../../linalg/trait.BaseMatrix.html) by removing the mean and scaling to unit variance.
+//!
+//! ### Usage Example
+//! ```
+//! use smartcore::api::{Transformer, UnsupervisedEstimator};
+//! use smartcore::linalg::naive::dense_matrix::DenseMatrix;
+//! use smartcore::preprocessing::numerical;
+//! let data = DenseMatrix::from_2d_vec(&vec![
+//!     vec![0.0, 0.0],
+//!     vec![0.0, 0.0],
+//!     vec![1.0, 1.0],
+//!     vec![1.0, 1.0],
+//! ]);
+//!
+//! let standard_scaler =
+//! numerical::StandardScaler::fit(&data, numerical::StandardScalerParameters::default())
+//!    .unwrap();
+//! let transformed_data = standard_scaler.transform(&data).unwrap();
+//! assert_eq!(
+//!     transformed_data,
+//!     DenseMatrix::from_2d_vec(&vec![
+//!         vec![-1.0, -1.0],
+//!         vec![-1.0, -1.0],
+//!         vec![1.0, 1.0],
+//!         vec![1.0, 1.0],
+//!     ])
+//! );
+//! ```
+use crate::api::{Transformer, UnsupervisedEstimator};
+use crate::error::{Failed, FailedError};
+use crate::linalg::Matrix;
+use crate::math::num::RealNumber;
+
+/// Configure Behaviour of `StandardScaler`.
+#[derive(Clone, Debug, Copy, Eq, PartialEq)]
+pub struct StandardScalerParameters {
+    /// Optionaly adjust mean to be zero.
+    with_mean: bool,
+    /// Optionally adjust standard-deviation to be one.
+    with_std: bool,
+}
+impl Default for StandardScalerParameters {
+    fn default() -> Self {
+        StandardScalerParameters {
+            with_mean: true,
+            with_std: true,
+        }
+    }
+}
+
+/// With the `StandardScaler` data can be adjusted so
+/// that every column has a mean of zero and a standard
+/// deviation of one. This can improve model training for
+/// scaling sensitive models like neural network or nearest
+/// neighbors based models.
+#[derive(Clone, Debug, Default, Eq, PartialEq)]
+pub struct StandardScaler<T: RealNumber> {
+    means: Vec<T>,
+    stds: Vec<T>,
+    parameters: StandardScalerParameters,
+}
+impl<T: RealNumber> StandardScaler<T> {
+    /// When the mean should be adjusted, the column mean
+    /// should be kept. Otherwise, replace it by zero.
+    fn adjust_column_mean(&self, mean: T) -> T {
+        if self.parameters.with_mean {
+            mean
+        } else {
+            T::zero()
+        }
+    }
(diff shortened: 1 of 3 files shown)


src/preprocessing/numerical.rs
```diff
@@ -0,0 +1,406 @@
...
+    /// should be kept. Otherwise, replace it by zero.
+    fn adjust_column_mean(&self, mean: T) -> T {
+        if self.parameters.with_mean {
+            mean
+        } else {
+            T::zero()
+        }
+    }
+    /// When the standard-deviation should be adjusted, the column
+    /// standard-deviation should be kept. Otherwise, replace it by zero.
```
Mec-iS, line 51: the docstring says "by zero" but it is instead substituted by one.

src/preprocessing/numerical.rs
```diff
@@ -0,0 +1,406 @@
+//! # Standard-Scaling For [RealNumber](../../math/num/trait.RealNumber.html) Matricies
+//! Transform a data [Matrix](../../linalg/trait.BaseMatrix.html) by removing the mean and scaling to unit variance.
+//!
+//! ### Usage Example
+//! ```
```
Mec-iS, line 5: Add the content of `fit_transform_with_default_standard_scaler` here using an example matrix

src/preprocessing/numerical.rs
```diff
@@ -0,0 +1,406 @@
...
+                        .div_scalar(self.adjust_column_std(*column_std))
+                })
+                .collect(),
+        )
+        .unwrap())
+    }
+}
+
+/// Extract a single column from a matrix and return it as a matrix.
+fn take_column_from_matrix<T, M>(matrix: &M, column_index: usize) -> M
```
Mec-iS, line 116: These two function look really useful, maybe you should consider adding them to `linalg::BaseMatrix` implementation as they are more general than the scaler implementation.
It would be nice to have something like `DenseMatrix::take_column(...)` (I thought we already have something like it?)

src/preprocessing/numerical.rs
```diff
@@ -0,0 +1,406 @@
...
+where
+    T: RealNumber,
+    M: Matrix<T>,
+{
+    matrix.take(&[column_index], 1)
+}
+
+/// From a collection of matrices, that contain columns, construct
+/// a matrix by stacking the columns horizontally.
+fn build_matrix_from_columns<T, M>(columns: Vec<M>) -> Option<M>
```
Mec-iS, line 126: Why this is a `Vec<M>` instead of a `Vec<Vec<T>>`>

Mec-iS (changes requested):
Thanks very much for your contribution!

These are the first suggetions that came to my mind.

src/preprocessing/numerical.rs
Mec-iS: there is already a `take` implementation for matrices.

src/preprocessing/numerical.rs
titoeb: Good catch! I will change it.

src/preprocessing/numerical.rs
titoeb: yet another good catch I totally forgot to add the example in the docstring, as I originally planned!

src/preprocessing/numerical.rs
titoeb: yes exactly, I also use the `Matrix::take` interface here, but I want to specialize it for taking columns (then you need to paramterize it via axis, e.g. `Matrix::take(&[column_index], 1)`). This is helpful because then I don't need to repeat this in my code and can test it in a central place (I always like to mess up the axis indices :laughing: ). So if it is ok, I would still like to keep `build_matrix_from_columns` (or put it in the `Matrix`-implementation).

Apart from this I would also be happy to put any of the two methods (`build_matrix_from_columns` | `take_column`) into the `Matrix` imlementation. Should I do it? For which of the two methods?

src/preprocessing/numerical.rs
titoeb: The `build_matrix_from_columns` function was build by me as a helper / utility function in this module. During `fit` of the `StandardScaler` I first take individual columns from the matrix that contains the raw data then apply the necessary scaling. In short I need `Vec<M>` instead of `Vec<Vec<T>>` because `take_column_from_matrix` returns a `Matrix`. Also this is helpful because I can use the neat methods `sub_scalar`  as well as `div_scaler` to apply the scaling.

src/preprocessing/numerical.rs
titoeb: I added an example, I did not use `fit_transform_with_default_standard_scaler` because that is just a small function I wrote for convenience in my testing suite, I did not plan to make it public to other user.

src/preprocessing/numerical.rs
Mec-iS: If they have generic functionalities and they can be reused, not strictly tied to the implementation of the module, they can be moved to modules with a more generic functionalities so that other modules can make use of them. I think you can move `take_columns_from_matrix` into `linalg::BaseMatrix` implementation so it can be used like `BaseMatrix::take_columns_from_matrix`; same for `build_matrix_from_columns`.

src/preprocessing/numerical.rs
Mec-iS: ok but for it to be general it is better to use `Vec<Vec<T>>` as you won't need build da matrix from a matrix but a matrix from a 2-D vector. So if you plan to make `build_matrix_from_columns` generic it would be better for it to accept Vecs

Mec-iS:
There is an error with clippy, try to run `cargo clippy --all-features -- -Drust-2018-idioms -Dwarnings`

src/preprocessing/numerical.rs
titoeb: Ok that make sense. Then I would prefer not to put `build_matrix_from_columns` into `Matrix`  because I would need to refactor it and it would not be optimal for this module anymore.

src/preprocessing/numerical.rs
titoeb: As we discussed below, I think `take_column_from_matrix` , or rather `Matrix:take_column` (what you proposed earlier) I would put it into a more central place as it can be used in other places as well.

titoeb:
> There is an error with clippy, try to run `cargo clippy --all-features -- -Drust-2018-idioms -Dwarnings`
Should be fixed now!

Mec-iS (approved):

Mec-iS:
Nice one @titoeb, hope we can collaborate in another PR soon.
If you are interested in how to serve `smartcore`, take a look to my repo https://github.com/Mec-iS/rust-grpc-smartcore

Issue #151, opened 2022-09-12
Implement multiclass classification metrics


montanalow:
The classification metrics don't currently support multiple classes. Other libraries use various techniques to generate the standard metrics like accuracy/precision/f1 etc by using averages across classes etc. e.g. from the [scikit precision metric](https://scikit-learn.org/stable/modules/generated/sklearn.metrics.precision_score.html#sklearn.metrics.precision_score)

```
average{‘micro’, ‘macro’, ‘samples’, ‘weighted’, ‘binary’} or None, default=’binary’
This parameter is required for multiclass/multilabel targets. If None, the scores for each class are returned. Otherwise, this determines the type of averaging performed on the data:

'binary':
Only report results for the class specified by pos_label. This is applicable only if targets (y_{true,pred}) are binary.

'micro':
Calculate metrics globally by counting the total true positives, false negatives and false positives.

'macro':
Calculate metrics for each label, and find their unweighted mean. This does not take label imbalance into account.

'weighted':
Calculate metrics for each label, and find their average weighted by support (the number of true instances for each label). This alters ‘macro’ to account for label imbalance; it can result in an F-score that is not between precision and recall.

'samples':
Calculate metrics for each instance, and find their average (only meaningful for multilabel classification where this differs from [accuracy_score](https://scikit-learn.org/stable/modules/generated/sklearn.metrics.accuracy_score.html#sklearn.metrics.accuracy_score)).
```

PR #152, opened 2022-09-13
Handle multiclass precision/recall

montanalow:
This implements `micro` averaging for multiclass precision/recall to handle #151. I believe this is a good default, since it handles imbalanced labels better than `macro` averaging (which is also not yet implemented).

Additionally, the documentation for `to_f32_bits` is incorrect, and that function may currently cause unexpected behavior because it truncates half the bits for f64 values. This breaks things like the uniqueness check I'm doing to count the number of distinct labels. I'd also be open to better methods to count distinct floating point values in Rust, but I haven't seen an easy library function that handles this (with nans etc). I've added `to_f64_bits` to provide the behavior I expected.

### DIFF
--- a/src/math/num.rs
+++ b/src/math/num.rs
@@ -46,8 +46,11 @@ pub trait RealNumber:
         self * self
     }
 
-    /// Raw transmutation to u64
+    /// Raw transmutation to u32
     fn to_f32_bits(self) -> u32;
+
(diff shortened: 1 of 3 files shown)


Mec-iS:
thanks for your contribution!

Mec-iS (approved):

Issue #153, opened 2022-09-14
Hyperparam tuning


montanalow:
It would be nice to have support for [hyperparam search](https://scikit-learn.org/stable/modules/generated/sklearn.model_selection.GridSearchCV.html#sklearn.model_selection.GridSearchCV) functionality. In addition to [KFold](https://docs.rs/smartcore/latest/smartcore/model_selection/index.html) cross validation. I'm planning to start work on a draft unless this is already underway.

PR #154, opened 2022-09-15
grid search

montanalow:
Implements a basic grid search for #153. This approach will requires that we implement Iterator SearchParameter structs for all estimators that have Parameters<T> that can be searched.

### DIFF
--- a/src/model_selection/hyper_tuning.rs
+++ b/src/model_selection/hyper_tuning.rs
@@ -1,0 +1,117 @@
+/// grid search results.
+#[derive(Clone, Debug)]
+pub struct GridSearchResult<T: RealNumber, I: Clone> {
+    /// Vector with test scores on each cv split
+    pub cross_validation_result: CrossValidationResult<T>,
+    /// Vector with training scores on each cv split
+    pub parameters: I,
+}
+
+/// Search for the best estimator by testing all possible combinations with cross-validation using given metric.
+/// * `fit_estimator` - a `fit` function of an estimator
+/// * `x` - features, matrix of size _NxM_ where _N_ is number of samples and _M_ is number of attributes.
+/// * `y` - target values, should be of size _N_
+/// * `parameter_search` - an iterator for parameters that will be tested.
+/// * `cv` - the cross-validation splitting strategy, should be an instance of [`BaseKFold`](./trait.BaseKFold.html)
+/// * `score` - a metric to use for evaluation, see [metrics](../metrics/index.html)
+pub fn grid_search<T, M, I, E, K, F, S>(
+    fit_estimator: F,
+    x: &M,
+    y: &M::RowVector,
+    parameter_search: I,
+    cv: K,
+    score: S,
+) -> Result<GridSearchResult<T, I::Item>, Failed>
+where
+    T: RealNumber,
+    M: Matrix<T>,
+    I: Iterator,
+    I::Item: Clone,
+    E: Predictor<M, M::RowVector>,
+    K: BaseKFold,
+    F: Fn(&M, &M::RowVector, I::Item) -> Result<E, Failed>,
+    S: Fn(&M::RowVector, &M::RowVector) -> T,
+{
+    let mut best_result: Option<CrossValidationResult<T>> = None;
+    let mut best_parameters = None;
+
+    for parameters in parameter_search {
+        let result = cross_validate(&fit_estimator, x, y, &parameters, &cv, &score)?;
+        if best_result.is_none()
+            || result.mean_test_score() > best_result.as_ref().unwrap().mean_test_score()
+        {
+            best_parameters = Some(parameters);
+            best_result = Some(result);
+        }
+    }
+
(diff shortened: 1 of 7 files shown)


Mec-iS:
This looks great!

* maybe it would be better to not use `mod.rs` but a dedicated `hyper_tuning.rs` module
* In general would be better to have more accurate assertions in the tests as far as possible, like asserting on values or arrays `assert_eq(result, expected)`. Also reusing [existing tests](https://scikit-learn.org/stable/auto_examples/model_selection/plot_grid_search_digits.html#tuning-hyper-parameters) to get expected values. 

If you have any idea about how to extend this (multiple metrics evaluation, randomized search), please open issues

montanalow:
I think it will be a similar sized effort (slightly larger) to implement randomized search with similar functionality to scikit, e.g. sampling from distributions. There might be an intermediate implementation if we only added a sampling Iterator that operates on the same parameter vectors, but instead of exhaustively enumerating every possible combination, it just samples from the grid. I don't think that interface would be as nice though, since it would put more work on the caller to pre-generate their arrays, although that is one of the options for scikits random search.

Mec-iS:
wow that's a lot of stuff! great work 👍🏼 

Let's try to merge this then we move on to what is next.

src/svm/svc.rs
```diff
@@ -102,6 +102,109 @@ pub struct SVCParameters<T: RealNumber, M: Matrix<T>, K: Kernel<T, M::RowVector>
...
+#[derive(Debug, Clone)]
+pub struct SVCSearchParameters<T: RealNumber, M: Matrix<T>, K: Kernel<T, M::RowVector>> {
+    /// Number of epochs.
+    pub epoch: Vec<usize>,
+    /// Regularization parameter.
+    pub c: Vec<T>,
+    /// Tolerance for stopping epoch.
+    pub tol: Vec<T>,
+    /// The kernel function.
+    pub kernel: Vec<K>,
```
montanalow, line 116: This won't work, since we may want to compare multi different types of kernels, and Vec must contain only a single type. SVM Kernel is the only example of this issue among the existing parameters. 

The options to fix don't appear straightforward to me. I haven't had success boxing the kernels, since they are templated themselves, and can't be `dyn`. I started looking at a potential `enum` based solution where we could store equal sized enum values that represent the kernels in a vec, but haven't gotten it working yet. 

I'm adding a test case to illustrate the issue.

levkk:
```rust
enum Kernels<T> {
  Linear,
  Polynomial(T, T, T),
}

fn main() {
  let v = vec![Kernels::<f32>::Linear, Kernels::<f32>::Polynomial(32.0, 1.0, 5.0)];
}
```

montanalow: I've started outlining a solution with the commit below that moves from the current `Kernel` implementations to an Enum version, but I believe this would be a breaking change.

src/model_selection/hyper_tuning.rs
```diff
@@ -0,0 +1,155 @@
...
+}
+
+/// Search for the best estimator by testing all possible combinations with cross-validation using given metric.
+/// * `fit_estimator` - a `fit` function of an estimator
+/// * `x` - features, matrix of size _NxM_ where _N_ is number of samples and _M_ is number of attributes.
+/// * `y` - target values, should be of size _N_
+/// * `parameter_search` - an iterator for parameters that will be tested.
+/// * `cv` - the cross-validation splitting strategy, should be an instance of [`BaseKFold`](./trait.BaseKFold.html)
+/// * `score` - a metric to use for evaluation, see [metrics](../metrics/index.html)
+pub fn grid_search<T, M, I, E, K, F, S>(
```
montanalow, line 23: There is a test case for this function, does this lint ignore tests?

montanalow (commented):
The test failure appears to be unrelated to the refactor at a surface level, but due to the new test case exercising SVC in more depth. I could use some help understanding the issue further since I'm less familiar with how SVC is actually implemented.

Mec-iS:
I think we grew this PR too much. We should rollback this one to [ebeb987](https://github.com/smartcorelib/smartcore/pull/154/commits/ebeb987e4852ca7279f989cf0e73804dda739ad5), merge those changes and open new PRs for the rest; otherwise everything can become hard to manage.

src/model_selection/hyper_tuning.rs
Mec-iS: change `svr.rs` line 717 to `let deserialized_svr: SVR<f64, DenseMatrix<f64>> = ...` and coverage will pass. No need to declare the last type.

montanalow:
As long as you are OK with merging while there are issues, breaking this up into smaller PRs works for me. Even though it's not complete, I think this is a good step forward.

Mec-iS:
I think it would be viable to add the tests that are failing and flag them with `#[ignore]` (so the checks pass but we keep track about what is failing).

Mec-iS (approved):

PR #157, opened 2022-09-16
SVC crashes on a test dataset with non default C

montanalow:
This test passes with a default C parameter of 1.0, but crashes on 0.0 or 10.0 at different places. It also passes with non default C values on trivial datasets, so it's an issue only exposed at larger scales.


Mec-iS:
this should be in a issue

PR #159, opened 2022-09-19
feat: allocate first and then proceed to create matrix from Vec of Ro…

morenol:
…wVectors

I was trying to use this function in a 5000 x 10000 matrix and it took a lot of time and it didn't finish, `v_stack` seems very expensive to be used here.

Not sure if we should return Option<> or an empty matrix when the vec is empty

### DIFF
--- a/src/linalg/mod.rs
+++ b/src/linalg/mod.rs
@@ -343,16 +343,16 @@ pub trait BaseMatrix<T: RealNumber>: Clone + Debug {
     ///    ])
     /// );
     fn from_row_vectors(rows: Vec<Self::RowVector>) -> Option<Self> {
-        if let Some(first_row) = rows.first().cloned() {
-            return Some(rows.iter().skip(1).cloned().fold(
-                Self::from_row_vector(first_row),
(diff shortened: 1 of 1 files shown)


morenol:
Any feedback is appreciated @titoeb

titoeb:
> Any feedback is appreciated @titoeb

Hey @morenol ! 
I am very happy you took this up. When I implemented the method, I did not bother too much with performance, but just tried to get the job done with the given interface. I can confirm that my implementation scales very poorly and your implementation works better.

With regard the the return value if the vector of row-vectors is empty we currently rely [on it](https://github.com/smartcorelib/smartcore/blob/436da104d7d6e96c43b58c106ade158ac9c5d446/src/readers/csv.rs#L91-L94). So therefore I would suggest keeping it for now. If there are some good reasons to change that behavior, that would also be ok for me, but then we also need to change the implementation I referenced.

I cannot review, but from my perspective, there is nothing blocking this PR.

Mec-iS:
Thanks people!

Mec-iS (approved):

PR #160, opened 2022-09-19
fix: fix compilation warnings when running only with default features

morenol:
I got:
```rust
warning: unused import: `itertools::Itertools`
 --> /Users/morenol/me/smartcore/src/algorithm/neighbour/fastpair.rs:2:5
  |
2 | use itertools::Itertools;
  |     ^^^^^^^^^^^^^^^^^^^^
  |
  = note: `#[warn(unused_imports)]` on by default
```
which is only used in force-brute implementation


Mec-iS:
thanks for this.

Mec-iS (approved):

Issue #165, opened 2022-09-21
Allow JSON construction of params


montanalow:
PostgresML instantiates algorithm params from JSON passed in at runtime. This isn't quite the same as serde deserialization, since the JSON params may be incomplete, and the missing ones will be filled in by Default::default(). We're currently adding this functionality downstream, but wondered if you'd be interested in adopting this upstream as well.

An example is in this PR https://github.com/postgresml/postgresml/pull/332/files#diff-a1bcda73e2901512a661eebf9e868be4f33dd99ac4952a85957397a860ec2035

montanalow:
I think it could be cleaner and easier to maintain here, by adopting `#[serde(default)]` on structs, if this functionality would be more broadly useful.

morenol:
> I think it could be cleaner and easier to maintain here, by adopting `#[serde(default)]` on structs, if this functionality would be more broadly useful.

I agree, I think that we can use `#[serde(default)]` 

I am in for this!

morenol:
Can we close this now, right? @montanalow

PR #166, opened 2022-09-21
Complete grid search params

montanalow:
Follow up to #154, that similarly adds iterable *SearchParameters for the remaining algorithms, so that they can also be used with `hyper_tuning::grid_search()`. 

- This also includes a cleanup to rename lasso_regression_search_parameters to lasso_search_parameters, that was misnamed in the original #154.
- Implement `Eq` for SVM Kernel structs to make them testable with assert_eq!


montanalow:
I wanted to get this done before #165, since all of these search params will also require `#[serde(default)]`.

src/svm/svc.rs
```diff
@@ -102,6 +102,109 @@ pub struct SVCParameters<T: RealNumber, M: Matrix<T>, K: Kernel<T, M::RowVector>
...
+#[derive(Debug, Clone)]
+pub struct SVCSearchParameters<T: RealNumber, M: Matrix<T>, K: Kernel<T, M::RowVector>> {
+    /// Number of epochs.
+    pub epoch: Vec<usize>,
+    /// Regularization parameter.
+    pub c: Vec<T>,
+    /// Tolerance for stopping epoch.
+    pub tol: Vec<T>,
+    /// The kernel function.
+    pub kernel: Vec<K>,
```
montanalow, line 116: This is the one param that it's currently impossible to meaningfully utilize, since we can't have multiple different K in the same Vec, but we can address that with a follow up PR.

Mec-iS:
ok just remember we will need your help to port this over to the new paradigm with 108 (;

Mec-iS (approved):

montanalow:
> ok just remember we will need your help to port this over to the new paradigm with 108 (;

I thought this would be orthogonal to the changes around arrays in 108, but happy to take a look if there is an issue.

PR #164, opened 2022-09-20
Lmm/add seeds in more algorithms

morenol:
Adds seeds parameter to multiple algorithms (including SVC) in order to fix the flaky test by using a fixed seed for that test.

Fixes #158

Also, adding the capability to disable `std` feature of `rand` crate so it can be used in WASM targets that don't have `getrandom` support.

Fixes #162

Adds a new `js` feature that can be enabled in wasm targets for explicit browser support.

PR #168, opened 2022-09-21
Add seed params to search params

montanalow:
#164 & #166 didn't have a direct conflict, but they did race while adding both seeds and search params to certain classes. This completes them together.

### DIFF
--- a/src/cluster/kmeans.rs
+++ b/src/cluster/kmeans.rs
@@ -145,6 +145,9 @@ pub struct KMeansSearchParameters {
     pub k: Vec<usize>,
     /// Maximum number of iterations of the k-means algorithm for a single run.
     pub max_iter: Vec<usize>,
+    /// Determines random number generation for centroid initialization.
+    /// Use an int to make the randomness deterministic
+    pub seed: Vec<Option<u64>>,
 }
 
 /// KMeans grid search iterator
@@ -152,6 +155,7 @@ pub struct KMeansSearchParametersIterator {
     kmeans_search_parameters: KMeansSearchParameters,
     current_k: usize,
     current_max_iter: usize,
+    current_seed: usize,
 }
 
 impl IntoIterator for KMeansSearchParameters {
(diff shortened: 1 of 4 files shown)


Mec-iS:
why tests did not spot these missing parts?

Mec-iS (approved):

montanalow:
As far as I can tell, this code doesn't even compile. I think tests pass independently on branchs, and the breakage only shows up when both branches are merged into master.

src/cluster/kmeans.rs
```diff
@@ -145,13 +145,17 @@ pub struct KMeansSearchParameters {
     pub k: Vec<usize>,
     /// Maximum number of iterations of the k-means algorithm for a single run.
     pub max_iter: Vec<usize>,
+    /// Determines random number generation for centroid initialization.
+    /// Use an int to make the randomness deterministic
+    pub seed: Vec<Option<u64>>,
```
morenol, line 150: Do we need Vec of Option? It looks to me that this should be either Option<u64> or Option<Vec<u64>>

montanalow: You're right, I think the cleanest type would be be Vec<u64>, although that'll require a little extra logic to convert to/from the Option<u64> seed type during iteration. I can take this up.

morenol: I am not completely sure if this is equivalent one-to-one, but in scikit it seems that random_state is only a number instead of a Vec, so not sure if we should try to match that here with a `Option<u64>` instead of the Vec

montanalow: I think you're looking at normal random_state params for estimators. Scikit grid search accepts a `Dict` with [values that are lists](https://scikit-learn.org/stable/modules/generated/sklearn.model_selection.GridSearchCV.html#sklearn.model_selection.GridSearchCV). 

In theory, I agree that it would seem odd to me that people would want to grid search with multiple params, but in practice I think that making this the only *SearchParameter that is scalar would be more surprising.

PR #169, opened 2022-09-22
Move to enum Kernels instead of Structs for hypertuning

montanalow:
SVCSearchParameters currently uses a Vec<K> for kernels, which means you can't search across multiple kernel types at the same time. This PR converts K to an enum Kernel type, instead of a generic type, so that multiple kinds of kernels may be compared in the same grid search.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Seeds to multiple algorithims that depend on random number generation.
 - Added feature `js` to use WASM in browser
 
-## BREAKING CHANGE
+## BREAKING CHANGES
+- SVM algorithms now use an enum type of Kernel, rather than distinct structs with a shared trait, so that they may be compared in a single hyperparameter search. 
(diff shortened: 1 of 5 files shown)


montanalow:
This is a breaking change, so may require more thought on the approach.

Mec-iS:
please add the breaking change to the CHANGESLOG

morenol:
I think that having the `Kernel` trait give us more flexibility. More people could implement their own Kernel and use it without troubles. I think that what we could do for ease of use is convert `Kernels` into an enum and implement `Kernel` for `Kernels`.

Mec-iS (changes requested):
Please adapt these changes to the new code in development.

`Kernel` is now an object-safe trait. Making an enum of `Kernel`s had been a problem for me because a match expression needs to return the same type. I tried different options but for sure there is something I am missing:
```
pun enum Kernels {
    Linear(LinearKernel),
    RBF(RBFKernel)
    ...
}

match kernel {
    Linear => LinearKernel::default(),  // this returns a LinearKernel type
    RBF => RBFkernel::default(),       // this returns a RBFKernel type
    ...
}
```

PR #173, opened 2022-09-23
Port num.rs from #108

Mec-iS:
I will add the "Authored" annotations at merge to assign the right credit to Volodymyr

Please provide a road map suggesting how to proceed next. Which modules are ready to be ported?


montanalow (commented):
I don't think I'm familiar enough with the usage of these primitives throughout SmartCore to effectively review and see issues that would arise.

Mec-iS:
I created a new branch where to proceed with the merging of the new code with the new num-traits implementation.

PR #179, opened 2022-10-01
feat: expose hyper tuning module in model_selection

morenol:
I just saw that this mod is not being imported so it is not being used and the tests are not running.

I noticed this while reviewing https://github.com/smartcorelib/smartcore/pull/169


morenol:
I will try to make API similar to https://scikit-learn.org/stable/modules/generated/sklearn.model_selection.GridSearchCV.html#sklearn.model_selection.GridSearchCV

Mec-iS (dismissed):

PR #203, opened 2022-10-29
Fix tests

morenol:
Fixes #

### Checklist

### DIFF
--- a/src/dataset/mod.rs
+++ b/src/dataset/mod.rs
@@ -8,9 +8,8 @@ pub mod digits;
 pub mod generator;
 pub mod iris;
 
-use crate::numbers::basenum::Number;
 #[cfg(not(target_arch = "wasm32"))]
-use crate::numbers::realnum::RealNumber;
+use crate::numbers::{basenum::Number, realnum::RealNumber};
 #[cfg(not(target_arch = "wasm32"))]
 use std::fs::File;
 use std::io;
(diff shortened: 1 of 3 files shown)


src/dataset/mod.rs
```diff
@@ -8,9 +8,8 @@ pub mod digits;
 pub mod generator;
 pub mod iris;
 
-use crate::numbers::basenum::Number;
 #[cfg(not(target_arch = "wasm32"))]
-use crate::numbers::realnum::RealNumber;
+use crate::numbers::{basenum::Number, realnum::RealNumber};
```
morenol, line 12: This is just a change to remove compilation warning about unused code in wasm32

src/naive_bayes/multinomial.rs
```diff
@@ -389,7 +389,6 @@ mod tests {
         assert!(iter.next().is_none());
     }
 
-    #[cfg_attr(target_arch = "wasm32", wasm_bindgen_test::wasm_bindgen_test)]
```
morenol, line 392: my bad, I have to add this again

src/linear/logistic_regression.rs
```diff
@@ -715,7 +715,7 @@ mod tests {
     #[cfg_attr(target_arch = "wasm32", wasm_bindgen_test::wasm_bindgen_test)]
     #[test]
     fn lr_fit_predict() {
-        let x = DenseMatrix::from_2d_array(&[
+        let x: DenseMatrix<f64> = DenseMatrix::from_2d_array(&[
```
morenol, line 718: using f64 instead of f32 for better precision in wasm32 tests

Mec-iS: probably f64 is more stable but possibly less supported

Issue #221, opened 2022-11-03
Deserialization for Kernel


Mec-iS:
`Kernel` is now an object-safe trait so we can use its type as `&dyn Kernel`, it has also a `name()` function to have a string representation of the kernel as `LinearKernel`, `RBFKernel`, ...

We would like to provide a functionality that instantiate the kernel's default from a string:
```
"LinearKernel" -> LinearKernel::Default()
```
via a enum
```
pub enum KernelName {
    Linear(LinearKernel),
   ...
}
```
This possibility is known to be not possible in vanilla Rust but there is a library that provides it, see [blogpost](https://fdeantoni.medium.com/rust-and-dynamically-loading-structs-c311d8edbef6).

This feature should be only available and loaded in pairing with the "serde" feature so to avoid adding another library to the deps.

This will allow parameters that contain `pub kernel: Option<&'a dyn Kernel<'a>>,` to be deserializable.

morenol:
I can take this, I think that we can get ride of `name` in Kernel trait, and instead handle that in the enum

Mec-iS:
ok, the problem with Rust comes when is needed to go string->instance. there is no easy way like in Python `getattr(module, string)` to read an object from a string.

PR #228, opened 2022-11-03
Use Box in SVM and remove lifetimes

morenol:
Fixes #

* Remove lifetimes in Kernel trait.
* Use box instead of & for Kernel objects n SVM structs.
* 
### Checklist

### DIFF
--- a/src/svm/svc.rs
+++ b/src/svm/svc.rs
@@ -57,8 +57,8 @@
 //! let y = vec![ -1, -1, -1, -1, -1, -1, -1, -1,
 //!            1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1];
 //!
-//! let knl = Kernels::linear();
-//! let params = &SVCParameters::default().with_c(200.0).with_kernel(&knl);
+//! let knl = Box::new(Kernels::linear());
+//! let params = &SVCParameters::default().with_c(200.0).with_kernel(knl);
 //! let svc = SVC::fit(&x, &y, params).unwrap();
 //!
 //! let y_hat = svc.predict(&x).unwrap();
@@ -91,15 +91,9 @@ use crate::rand_custom::get_rng_impl;
 use crate::svm::Kernel;
 
 #[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
-#[derive(Debug, Clone)]
+#[derive(Debug)]
 /// SVC Parameters
-pub struct SVCParameters<
-    'a,
-    TX: Number + RealNumber,
-    TY: Number + Ord,
-    X: Array2<TX>,
-    Y: Array1<TY>,
-> {
+pub struct SVCParameters<TX: Number + RealNumber, TY: Number + Ord, X: Array2<TX>, Y: Array1<TY>> {
     /// Number of epochs.
     pub epoch: usize,
     /// Regularization parameter.
@@ -108,7 +102,7 @@ pub struct SVCParameters<
     pub tol: TX,
     #[cfg_attr(feature = "serde", serde(skip_deserializing))]
     /// The kernel function.
-    pub kernel: Option<&'a dyn Kernel<'a>>,
+    pub kernel: Option<Box<dyn Kernel>>,
     /// Unused parameter.
     m: PhantomData<(X, Y, TY)>,
     /// Controls the pseudo random number generation for shuffling the data for probability estimates
@@ -129,7 +123,7 @@ pub struct SVC<'a, TX: Number + RealNumber, TY: Number + Ord, X: Array2<TX>, Y: 
     classes: Option<Vec<TY>>,
     instances: Option<Vec<Vec<TX>>>,
     #[cfg_attr(feature = "serde", serde(skip))]
-    parameters: Option<&'a SVCParameters<'a, TX, TY, X, Y>>,
+    parameters: Option<&'a SVCParameters<TX, TY, X, Y>>,
     w: Option<Vec<TX>>,
     b: Option<TX>,
     phantomdata: PhantomData<(X, Y)>,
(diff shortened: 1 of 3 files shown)


Mec-iS:
i like the idea of making the kernels name static but we cannot use Box for the kernel as it ruins the API call. For an example of how kernels are used see the Jupyter notebooks:
```
let y_hat: Vec<f64> = SVC::fit(
    &x, &y,
    &SVCParameters::default()
        .with_c(1.0)
        .with_kernel(&Kernels::rbf()
            .with_gamma(0.7)),
     ) .and_then(
        |lr| lr.predict(&x)
) .unwrap(); 

```
how the kernels are used in tests is irrelevant. we don't want the end-users to allocate to the heap manually.
Kernels are intended to be used as references in parameters and it is made possible by them being object-safe (`dyn...`). As far as I understood, for object-safe trait using & or using box is the same as they are dynamically allocated at runtime, see https://www.reddit.com/r/rust/comments/kw8p5v/what_does_it_mean_to_be_an_objectsafe_trait/ or "Advanced Traits" in the Rust book.

Mec-iS:
about implementing an enum for kernels, see https://github.com/smartcorelib/smartcore/issues/221

src/svm/svc.rs
```diff
@@ -108,7 +102,7 @@ pub struct SVCParameters<
     pub tol: TX,
     #[cfg_attr(feature = "serde", serde(skip_deserializing))]
     /// The kernel function.
-    pub kernel: Option<&'a dyn Kernel<'a>>,
+    pub kernel: Option<Box<dyn Kernel>>,
```
Mec-iS, line 105: this one is the same as writing `Option<&dyn Kernel>` as both are allocated at runtime, why change it>

src/svm/svc.rs
```diff
@@ -184,8 +178,8 @@ impl<'a, TX: Number + RealNumber, TY: Number + Ord, X: Array2<TX>, Y: Array1<TY>
         self
     }
     /// The kernel function.
-    pub fn with_kernel(mut self, kernel: &'a (dyn Kernel<'a>)) -> Self {
-        self.kernel = Some(kernel);
+    pub fn with_kernel<K: Kernel + 'static>(mut self, kernel: K) -> Self {
```
Mec-iS, line 181: this makes the kernel object to live for the entire execution of the program: https://doc.rust-lang.org/rust-by-example/scope/lifetime/static_lifetime.html#reference-lifetime

src/svm/svc.rs
```diff
@@ -184,8 +178,8 @@ impl<'a, TX: Number + RealNumber, TY: Number + Ord, X: Array2<TX>, Y: Array1<TY>
...
+        self.kernel = Some(Box::new(kernel));
```
Mec-iS, line 182: same as above, as Kernel is object-safe, same as writing `Some(&dyn Kernel)`

Mec-iS:
be careful. I agreed in using `'static` for the *strings* used by kernels' names, but if you implement the whole Kernel as `'static` itself it means that you are keeping in memory every kernel instantiated for the entire length of the program.

src/svm/svc.rs
```diff
@@ -251,7 +245,7 @@ impl<'a, TX: Number + RealNumber, TY: Number + Ord, X: Array2<TX> + 'a, Y: Array
     pub fn fit(
         x: &'a X,
         y: &'a Y,
-        parameters: &'a SVCParameters<'a, TX, TY, X, Y>,
+        parameters: &'a SVCParameters<TX, TY, X, Y>,
```
Mec-iS, line 248: this change looks right, but we can be sure that all the attributes inside `SVCParameters` have the same lifetime of `&'a`? No, because Kernel has no lifetime (aka its lifetime is `'static` )

src/svm/svc.rs
morenol: I think that at this case it means this: https://doc.rust-lang.org/rust-by-example/scope/lifetime/static_lifetime.html#trait-bound.

```As a trait bound, it means the type does not contain any non-static references```

src/svm/svc.rs
morenol: I thik that `'static` in this case has a different meaning than 'static lifetime

src/svm/svc.rs
morenol: just to get ride of lifetimes associated to the type

src/svm/svc.rs
Mec-iS: Ah right. lifetime have different syntax than traits, for which can be `!Sized` or also `?Sized`.

Ok then if we are sure that the kernel has right lifetime (e.g. is dropped when assigned variable is out of scope) this can work as expected.

Mec-iS:
@morenol Ok then, please keep the `&dyn Kernel` syntax for type declarations as it is the preferred one since Rust 2021.
The rest is OK.

Good one!

src/svm/svc.rs
Mec-iS: ok, but why using `Box`? `pub kernel: Option<&dyn Kernel>,` should do and it is more inline with Rust 2021

morenol: I think that compiler will complain in  `Option<&dyn Kernel>,`, in that case I think that it is needed a explicit lifetime definition

morenol: I will try that in any case. Could you share with me a reference to the Rust 2021 suggestion?

Mec-iS: https://doc.rust-lang.org/edition-guide/rust-2018/new-keywords.html

Mec-iS (approved):

src/svm/svc.rs
Mec-iS: I prefer to use `Option<&dyn ...>` because is easier to access: `variable.as_ref().unwrap()` (if you don't want to move) instead of `*variable.as_ref().unwrap()`. Are there easier ways to borrow the content of a box without moving it?

morenol: I think that we can use `deref()`

src/svm/svc.rs
Mec-iS, line 181: isn't better here to pass a reference? `...(mut self, kernel: &K))`

PR #231, opened 2022-11-04
refactor: do not require an empty struct in cross_validate to define …

### LINKED ISSUE #230: Use Default::default instead of new in trait
### I'm submitting a 
- [x] improvement.

### Current Behaviour:

Currently, traits in https://github.com/smartcorelib/smartcore/blob/development/src/api.rs#L24 requires implementation of  `fn new() -> Self;`, I think that instead we should rely on Default::default,

Yes

Any thoughts?

morenol:
…which is the estimator

Fixes https://github.com/smartcorelib/smartcore/issues/230

### Checklist

### DIFF
--- a/src/api.rs
+++ b/src/api.rs
@@ -18,10 +18,6 @@ pub trait UnsupervisedEstimator<X, P> {
 
 /// An estimator for supervised learning, that provides method `fit` to learn from data and training values
 pub trait SupervisedEstimator<X, Y, P>: Predictor<X, Y> {
-    /// Empty constructor, instantiate an empty estimator. Object is dropped as soon as `fit()` is called.
-    /// used to pass around the correct `fit()` implementation.
-    /// by calling `::fit()`. mostly used to be used with `model_selection::cross_validate(...)`
-    fn new() -> Self;
     /// Fit a model to a training dataset, estimate model's parameters.
     /// * `x` - _NxM_ matrix with _N_ observations and _M_ features in each observation.
     /// * `y` - target training values of size _N_.
@@ -36,10 +32,6 @@ pub trait SupervisedEstimator<X, Y, P>: Predictor<X, Y> {
 /// In this one parameters are borrowed instead of moved, this is useful for parameters that carry
 /// references. Also to be used when there is no predictor attached to the estimator.
 pub trait SupervisedEstimatorBorrow<'a, X, Y, P> {
-    /// Empty constructor, instantiate an empty estimator. Object is dropped as soon as `fit()` is called.
-    /// used to pass around the correct `fit()` implementation.
-    /// by calling `::fit()`. mostly used to be used with `model_selection::cross_validate(...)`
-    fn new() -> Self;
     /// Fit a model to a training dataset, estimate model's parameters.
     /// * `x` - _NxM_ matrix with _N_ observations and _M_ features in each observation.
     /// * `y` - target training values of size _N_.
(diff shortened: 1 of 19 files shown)


morenol:
Even though, it looks like more easy to use this passing the parameter, it looks weird to me to pass an unitialized empty struct, instead we can take advantage of that we can explicitly define which is the Estimator that we want to use

Mec-iS:
@morenol from a programming perspective I agree, but writing `::<_, _, _, _, _, LogisticRegression<_, _, _, _>, _, _>` is really not viable for an interface that has to face an end-user. Think if you had to write that to issue a request using Python `requests`: `requests.request::<_,_,_,_, POST,_,_,_>(...)` ?!!? would you ever use something like that?

If you want to remove the `new` there are two option imo:
* passing the single `fit` function using `Fn` or `FnOnce` but it is hard as every estimator has different signatures
* passing the string name `"LogisticRegression"` and instantiate an object inside `fit` using an enumerator (as proposed for Kernel in #221) or an enumerator key like `EstimatorEnum::LogisticRegression`
* refactoring `cross_validate` in a way that the user has to write only `cross_validate::<LogisticRegression>(...)`
* anything else I don't know about

I used `new` because there are some other constraints in place and the alternatives are not straightforward. As we have as a main point to keep an easy "pythonic" API we cannot allow Rust's types to reach the end-user's level.

morenol:
> @morenol from a programming perspective I agree, but writing `::<_, _, _, _, _, LogisticRegression<_, _, _, _>, _, _>` is really not viable for an interface that has to face an end-user. Think if you had to write that to issue a request using Python `requests`: `requests.request::<_,_,_,_, POST,_,_,_>(...)` ?!!? would you ever use something like that?
> 
> If you want to remove the `new` there are these options imo:
> 
> * passing the single `fit` function using `Fn` or `FnOnce` but it is hard as every estimator has different signatures
> * passing the string name `"LogisticRegression"` and instantiate an object inside `fit` using an enumerator (as proposed for Kernel in [Deserialization for Kernel #221](https://github.com/smartcorelib/smartcore/issues/221)) or an enumerator key like `EstimatorEnum::LogisticRegression`
> * refactoring `cross_validate` in a way that the user has to write only `cross_validate::<LogisticRegression>(...)`
> * anything else I don't know about
> 
> I used `new` because there are some other constraints in place and the alternatives are not straightforward. As we have as a main point to keep an easy "pythonic" API we cannot allow Rust's types to reach the end-user's level.
> 
> PS: I don't like how `cross_validate` and `cross_validate_predict` are implemented, so maybe we should rewrite them in a better way. This could also lead to a better definition of the `SupervisedEstimator` and `Predictor` traits. For an example see [fit](https://rust-ml.github.io/linfa/rustdocs/linfa/traits/trait.Fit.html) and [predict](https://rust-ml.github.io/linfa/rustdocs/linfa/traits/trait.Predict.html).

I agree, I am not fan of having an API were something like this is required `::<_, _, _, _, _, LogisticRegression<_, _, _, _>, _, _>`. I will keep this open until I find a better way

Sammyreal1:
> Even though, it looks like more easy to use this passing the parameter, it looks weird to me to pass an unitialized empty struct, instead we can take advantage of that we can explicitly define which is the Estimator that we want to use using turbofish syntax

Mec-iS:
> > Even though, it looks like more easy to use this passing the parameter, it looks weird to me to pass an unitialized empty struct, instead we can take advantage of that we can explicitly define which is the Estimator that we want to use using turbofish syntax

see discussion at https://github.com/smartcorelib/smartcore/issues/230

PR #232, opened 2022-11-04
Handle kernel serialization

### LINKED ISSUE #221: Deserialization for Kernel

morenol:
Fixes https://github.com/smartcorelib/smartcore/issues/221

### Checklist

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -35,7 +35,7 @@ typetag = { version = "0.2", optional = true }
 
 [features]
 default = []
-serde = ["dep:serde"]
+serde = ["dep:serde", "dep:typetag"]
 ndarray-bindings = ["dep:ndarray"]
 datasets = ["dep:rand_distr", "std_rand", "serde"]
 std_rand = ["rand/std_rng", "rand/std"]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -21,10 +21,11 @@ num = "0.4"
 rand = { version = "0.8.5", default-features = false, features = ["small_rng"] }
 rand_distr = { version = "0.4", optional = true }
 serde = { version = "1", features = ["derive"], optional = true }
+typetag = { version = "0.2", optional = true }
 
 [features]
 default = ["serde", "datasets"]
-serde = ["dep:serde"]
+serde = ["dep:serde", "dep:typetag"]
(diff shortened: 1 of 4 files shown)


morenol:
Wow, typetag is not usable in wasm, I disabled that for WASM32 target. So we have full serialization/deserialization support in all targets except WASM

Mec-iS:
that is why I wanted default to be empty feature:
`default = []`
 so that it could compiled to Wasm, we can assume that for in-browser application serialisation is not needed in the Wasm binary.

please add some tests for Kernel serialisation/deserialisation

morenol:
> that is why I wanted default to be empty feature: `default = []` so that it could compiled to Wasm, we can assume that for in-browser application serialisation is not needed in the Wasm binary.
> 
> please add some tests for Kernel serialisation/deserialisation

Addressed.

I agree. I think that by default we should not enable any feature

morenol:
> so that it could compiled to Wasm, we can assume that for in-browser application serialisation is not needed in the Wasm binary.

I think that there could be a couple of use cases in wasm where deserialization could be useful, but I think that is ok for now to partially support WASM

Mec-iS:
> I think that there could be a couple of use cases in wasm where deserialization could be useful, but I think that is ok for now to partially support WASM

yes something we can add later or leave it to the implementation in JS

Mec-iS:
@morenol please double check that these changes are inline with the new features setup.

Mec-iS (approved):

Issue #214, opened 2022-11-01
Randomness: implementation and use getrandom for Wasm target


Mec-iS:
### I'm submitting a 
- [x] feature request.

### Current Behaviour:
`RealNumber::rand` is not implemented. `Floatnumber::rand` use an external crate that cannot work on Wasm. 

the use of random in `fn rand()` is ambiguous and may fail in some targets. Some decisions have to be taken about the usage of `rand::Rng` also. I suggest going for using `getrandom` as it applies to the most [targets](https://docs.rs/getrandom/latest/getrandom/).

cc: @morenol

Mec-iS:
related to #216

morenol:
We use `getrandom` with `js` feature for WASM in browser target:

https://github.com/smartcorelib/smartcore/blob/4cf7e4d7b7834f26ec3d8b3a37a136e84ad842ca/Cargo.toml#L32

I think that `rand` works without troubles for WASI target

Mec-iS:
ok then we need to clear up the features:
```
default = ["serde", "wasi", "datasets"]  // datasets should also use gerandom
wasi = [..add deps ..]
...
```
optimally for wasi we should be able to do also `no_std`?

morenol:
I think that we can get ride of that issue using

```
        let mut rng = get_rng_impl(None);
```

instead of 
```
let mut rng = rand::thread_rng();
```

Mec-iS:
ok but then shouldn't we use `getrandom` instead of `rand` for the reasons above?
`get_rng_impl(None)` uses `rand`.

morenol:
Not completely sure if we need this. Rand is the high level API and the README of that project says: 

```
This is a low-level API. Most users should prefer using high-level random-number library like [rand](https://crates.io/crates/rand).
```

Mec-iS:
we need lower-level implementation for default feature

Mec-iS:
This is not fixed yet

PR #235, opened 2022-11-07
Release 0.3

Mec-iS:
This is the release candidate for v0.3

Please take a look, run some tests, etc.

Changes compared to `development`:
* improve documentation
* ~make `default` feature empty (no optional dependencies)~ refactor feature system 
* cleanup unused variables

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -16,6 +16,7 @@ exclude = [
     ".gitignore",
     "smartcore.iml",
     "smartcore.svg",
+    "tests/"
 ]
 
 [dependencies]
@@ -25,6 +26,7 @@ ndarray = { version = "0.15", optional = true }
 num-traits = "0.2.12"
 num = "0.4"
 rand = { version = "0.8.5", default-features = false, features = ["small_rng"] }
+getrandom = { version = "*", features = ["js"] }
 rand_distr = { version = "0.4", optional = true }
 serde = { version = "1", features = ["derive"], optional = true }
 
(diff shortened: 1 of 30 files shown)


Mec-iS:
@morenol please take a look

we have to fix the complications with the random numbers range generator at  https://github.com/smartcorelib/smartcore/issues/214

we have now a clearer setup for the features:
* `default`:  only base dependencies (no optional), should work with Wasi
* `serde`: whatever requires serialization
* `datasets`: whatever requires datasets
* `std`
* other

I don't understand the role of the `std` feature, I suppose is meant to use the standard library randomization library.
Now, we have to assure that:
* Wasm/Wasi (default) use the right implementation for random (currently getrandom), 
* std feature uses the standard random
* datasets is currently using `rand_distr`, is there the possibility to make it use std random or getrandom?

Mec-iS:
Fixed!

Now features are now:
```
[features]
default = []
serde = ["dep:serde"]
ndarray-bindings = ["dep:ndarray"]
datasets = ["dep:rand_distr", "std_rand", "serde"]
std_rand = ["rand/std_rng", "rand/std"]
```

src/cluster/kmeans.rs
```diff
@@ -74,7 +74,7 @@ pub struct KMeans<TX: Number, TY: Number, X: Array2<TX>, Y: Array1<TY>> {
     k: usize,
     _y: Vec<usize>,
     size: Vec<usize>,
-    distortion: f64,
+    _distortion: f64,
```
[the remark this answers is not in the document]
Mec-iS: let's keep it for now, they are just single-valued

Cargo.toml
```diff
@@ -19,32 +26,29 @@ ndarray = { version = "0.15", optional = true }
 num-traits = "0.2.12"
 num = "0.4"
 rand = { version = "0.8.5", default-features = false, features = ["small_rng"] }
+getrandom = { version = "*", features = ["js"] }
```
morenol, line 29: Let me check if there are issues with enabling `js` always.

Mec-iS: do we need it right?
I think it is just an additional feature, shouldn't affect the base functionality

Mec-iS: we need an alternative for [this branch of the match](https://github.com/smartcorelib/smartcore/pull/235/files#diff-f66732050b1be4b9113c09c195087cad63579455d9f708ec0f97eacfebb3fc68R17)

morenol: So, this feature flag only has effect for `wasm32-unknown-unknown`. Enabling this feature breakes the usage of wasm32-unknown-unknown with wasmtime and wasmer because it adds wasmbindgen to the build

morenol: We can avoid that branch by setting a seed for those targets. In any case I think that we could use get_random but only enabling "js" feature if it is a explicitly enabled feature

morenol: I mean,  I think that we should keep this line in the features section:

`
js = ["getrandom/js"]
`

and remove js from being enabled by default

Cargo.toml
```diff
@@ -36,6 +36,8 @@ serde = ["dep:serde"]
 ndarray-bindings = ["dep:ndarray"]
 datasets = ["dep:rand_distr", "std_rand", "serde"]
 std_rand = ["rand/std_rng", "rand/std"]
+# used by wasm32-unknown-unknown
```
morenol, line 39: Maybe we could add here in the comment that is using by wasm32-unknown-unknown when we want to target browser

Cargo.toml
morenol: my bad it seems to not be working with this change

morenol: Can we move all the changes related to rand vs getrandom to another PR? so we can merge the other changes?

morenol: @Mec-iS I think that we shouldnt merge this PR

PR #236, opened 2022-11-08
Ready for Release 0.3.0

Mec-iS:
merge to `main`


morenol:
@Mec-iS I think that we should fix some issues with getrandom in wasm32 when we want to target non-browser targets

Mec-iS:
ok. which issuea, please be specific.

morenol:
`cargo build --target wasm32-unknown-unknown` should compile with success without the need to enable the js feature

morenol:
Also, should we merge https://github.com/smartcorelib/smartcore/pull/232 before release?

Cargo.toml
```diff
@@ -40,7 +40,7 @@ std_rand = ["rand/std_rng", "rand/std"]
 js = ["getrandom/js"]
 
 [target.'cfg(target_arch = "wasm32")'.dependencies]
-getrandom = { version = "0.2", optional = true }
+getrandom = { version = "*", features = ["js"] }
```
morenol, line 43: this enforces the usage of js feature for wasm32 target, will fix cargo. build  but will produce builds only supported by wasm32,  I did this PR https://github.com/smartcorelib/smartcore/pull/237/files and seems to be working. 

I have to say that getrandom for wasm32-unknown-unknown assumes that the target is the browser, for that reason I am only using that if we use the js feature

morenol: There are a couple of projects that I know that uses wasm32-unknown outside of the browser. I think that if we can support that use case with something like https://github.com/smartcorelib/smartcore/pull/237/files we should proceed with that

```
wasm32-unknown-unknown - this target, like the WASI one, is focused on producing single *.wasm binaries. The standard library, however, is largely stubbed out since the "unknown" part of the target means libstd can't assume anything. This means that while binaries will likely work in wasmtime, common conveniences like println! or panic! won't work.
```

https://docs.wasmtime.dev/wasm-rust.html

morenol (approved):
LGTM, thanks for leading all the effort to accomplish this. Other than maybe merging this first, https://github.com/smartcorelib/smartcore/pull/232 it looks good to me

Mec-iS:
yes we merge 232 first

morenol (approved):

PR #237, opened 2022-11-08
Fixes for release

morenol:
Fixes #

### Checklist


morenol:
> I think you are overthinking it. It looks quite easy to me:
> 
> * `wasm32-unknown` is for browsers
> * `wasm32-wasi` is for other VMs that have access to binary interfaces
> 
> Also, let's try to not duplicate pipelines in tests, once we run `cargo tests --all-features` for each target isn't it enough?

In [Fluvio](https://github.com/infinyon/fluvio) we use wasm32-unknown-unknown with wasmtime. I think that should not assume that thing.

morenol:
We recently added support for WASI, but still is not enabled by default, we could consider to migrate to only use WASI in the feature but that decission has not been taken yet

PR #251, opened 2023-03-20
Implement realnum::rand

### LINKED ISSUE #214: Randomness: implementation and use getrandom for Wasm target

Mec-iS:
Implement rand. Use the new derive [#default]

Fixes #214 

### Checklist

### Current behaviour
`realnum::rand` returns `1.0`
### New expected behaviour
random number is generated


src/numbers/realnum.rs
```diff
@@ -63,8 +66,12 @@ impl RealNumber for f64 {
     }
 
     fn rand() -> f64 {
-        // TODO: to be implemented, see issue smartcore#214
-        1.0
+        let mut small_rng = SmallRng::from_seed(SEEDS);
```
morenol: could you call https://github.com/smartcorelib/smartcore/blob/development/src/rand_custom.rs here? so it can compile for wasm targets? both browser and non-browser tagets

Mec-iS:
thanks @morenol
I need a `0.3.1` release asap! 👍🏼

src/numbers/realnum.rs
```diff
@@ -66,7 +67,7 @@ impl RealNumber for f64 {
     }
 
     fn rand() -> f64 {
-        let mut small_rng = SmallRng::from_seed(SEEDS);
+        let mut small_rng = get_rng_impl(Some(100101010));
```
morenol: I think that if we pass `None` here, we will get a random seed for std targets. We could change `get_rng_impl` to get a seed from current time for other targets maybe?

morenol:
> thanks @morenol I need a `0.3.1` release asap! 👍🏼

Can you bump the version on this PR?

src/cluster/dbscan.rs
```diff
@@ -18,7 +18,7 @@
 //!
 //! Example:
 //!
-//! ```
+//! ```ignore
```
morenol, line 21: Why is this needed?

morenol (dismissed):
LGTM but not sure why `ignore` was added to one of the doc tests

Mec-iS:
> LGTM but not sure why `ignore` was added to one of the doc tests

that test uses `datasets` feature so it fails, it is replicated among the `tests` with the right flag.

I may have more changes in the next hours, I am adding anti-panicking checks to `fit` methods.

morenol:
> 
> I may have more changes in the next hours, I am adding anti-panicking checks to `fit` methods.

I like that idea!

morenol:
It seems that there is conflicts, it says to me that there are 83 commits on this PR, could you rebase?

Mec-iS:
>It seems that there is conflicts, it says to me that there are 83 commits on this PR, could you rebase?

that is because I pushed the wrong branch at the creation of the PR, then I corrected it.
Check in "Files Changed" that the changes are only the latest ones.

EDIT: your local branch may be missing the "Update README.md" I pushed before this PR.

Issue #256, opened 2023-04-02
Show clear input-shape validation errors instead of index out of bounds panics (was: index out of bounds using logistic_regression)


EdmundsEcho:
### I'm submitting a 
- [ x] bug report.
- [ ] improvement.
- [ ] feature request.

### Current Behaviour:

When I call
```rust
 let lr = LogisticRegression::fit(x, y, Default::default())?;
```

I get an index out of bounds error.
```fish
thread 'main' panicked at 'index out of bounds: the len is 4852833 but the index is 4852833'
```

### Expected Behaviour:

It should compute the linear regression without throwing an error.

### Steps to reproduce:
These are the input shapes
```rust
y vec height: 52181  // Vec<u32>
x matrix shape: (52181, 94) // DenseMatrix<f64>
```

### Environment:
 
* rustc version: rustc 1.68.2 (9eb3afe9e 2023-03-27)
* cargo version: cargo 1.68.2 (6feb7c9cf 2023-03-26)
* OS details: MacOS 13.3 M1

### Do you want to work on this issue?
yes.  I'm trying to debug it now.  This said, I'm new to this lib and so want to make sure I'm not missing something.

EdmundsEcho:
## Update
The error happens when running the optimization, calling `get` trait method for `DenseMatrix`.
The optimization/first_order/lbfgs.rs

```rust
    fn optimize<'a, X: Array1<T>, LS: LineSearchMethod<T>>(
        &self,
        f: &F<'_, T, X>,
        df: &'a DF<'_, X>,
        x0: &X,
        ls: &'a LS,
    ) -> OptimizerResult<T, X> {
        let mut state = self.init_state(x0);

        df(&mut state.x_df, x0);  // <<< 

     ...
}
```

Mec-iS:
Thanks for reporting this.
I will also take some time to look into it, there should be a test that covers LR, please add a test for this if you find the problem.

Mec-iS:
Hi @EdmundsEcho
I don't know how could you even make your code to run actually. Please check this:
* are you using the latest version `>=0.3.0`? if not, please use the latest version as it is the only API supported
* check the examples and tests in `src/linear/logistic_regression.rs` line 858 for examples about how to use the method

there are two major problems with your method call  `let lr = LogisticRegression::fit(x, y, Default::default())?;`:

[...]

PR #257, opened 2023-04-03
Patch to version 0.4.0

Mec-iS:
Collecting changes for version `0.4.0`

### New expected behaviour
see #256 

### Change logs
* Add Logistic Regression random test
* Add `Result` type to `DenseMatrix` constructor by @EdmundsEcho
* BREAKING CHANGES on `DenseMatrix` constructor, now constructor returns a `Result`


morenol (dismissed):
LGTM

Mec-iS:
@morenol we need to fix Wasm tests

src/linalg/basic/matrix.rs
```diff
@@ -49,27 +51,75 @@ pub struct DenseMatrixMutView<'a, T: Debug + Display + Copy + Sized> {
     column_major: bool,
 }
 
+// functional utility functions used across types
```
Mec-iS: @EdmundsEcho
Please follow Rust code conventions, every functions shall be documented using triple slashes `///`. I know this is boring and nitpicking but we have to stick to conventions. Also the content of the documentation shall be explicative of what the function does; "functional utilities" is very generic.

src/linalg/basic/matrix.rs
```diff
@@ -49,27 +51,75 @@ pub struct DenseMatrixMutView<'a, T: Debug + Display + Copy + Sized> {
...
+    );
+    debug_assert!(
+        // depends on a properly formed range
+        vrows.start <= vrows.end && vcols.start <= vcols.end,
+        "Invalid range: start <= end failed"
+    );
+
+    !(vrows.end <= mrows && vcols.end <= mcols && vrows.start <= mrows && vcols.start <= mcols)
+}
+fn start_end_stride(
```
Mec-iS: missing documentation for this function

src/linalg/basic/matrix.rs
```diff
@@ -182,36 +233,65 @@ impl<'a, T: Debug + Display + Copy + Sized> fmt::Display for DenseMatrixMutView<
...
+            !(values.is_empty() || values[0].is_empty()),
+            "Instantiating DenseMatrix requires a non-empty 2d_vec"
+        );
+
+        if values.is_empty() || values[0].is_empty() {
+            Err(Failed::input(
+                "The 2d vec provided is empty; cannot instantiate the matrix",
+            ))
+        } else {
+            let nrows = values.len();
```
Mec-iS, line 221: this is inefficient, why would you make a copy of the values into `m_values` when you already have them in memory passed as a reference? Smartcore aims to minimize copying.
All these changes will make instantiation more costly and slower. 

I see also some redundancy, why are doing: 
```
                values
                .first()
                .unwrap_or_else(|| {
                    panic!("Invalid state: Cannot create 2d matrix from an empty vector")
```
when we already checked  `if values.is_empty() || values[0].is_empty()` above?

The `else` should just be:
```
else {
    DenseMatrix::from_2d_vec(&values.iter().map(|row| Vec::from(*row)).collect())
}
```
Please be considerate, adding useless for-loops is never a good choice.

src/linalg/basic/matrix.rs
EdmundsEcho: I can describe each. I did not think it should be part of the public documentation as they are "internal/private".  So, `//` or `///`? What do you think?

Mec-iS: they should probably part of a trait, but I dont know which one, or a new.
Are they general or specific to a struct?

src/linalg/basic/matrix.rs
Mec-iS: Sorry this is wrong. it is like this also in the current version.

src/linalg/basic/matrix.rs
```diff
@@ -544,15 +627,74 @@ mod tests {
...
+        assert!(x.is_err());
+    }
+    #[test]
+    fn test_instantiate_from_2d_empty2() {
+        let input: &[&[f64]] = &[&[], &[]];
+        let x = DenseMatrix::from_2d_array(input);
+        assert!(x.is_err());
+    }
+    #[test]
+    fn test_instantiate_ok_view1() {
```
Mec-iS, line 629: These tests are failing:
```
test linalg::basic::matrix::tests::test_instantiate_err_view1 ... FAILED
test linalg::basic::matrix::tests::test_instantiate_err_view2 ... FAILED
test linalg::basic::matrix::tests::test_instantiate_err_view3 ... FAILED
test linalg::basic::matrix::tests::test_instantiate_from_2d ... ok
test linalg::basic::matrix::tests::test_instantiate_from_2d_empty ... FAILED
test linalg::basic::matrix::tests::test_instantiate_from_2d_empty2 ... FAILED
```

src/linalg/basic/matrix.rs
Mec-iS: ...or maybe they should just be part of the structure `impl` (as "static" methods) if they are used only by this particular structure

src/linalg/basic/matrix.rs
Mec-iS:
```
thread 'linalg::basic::matrix::tests::test_instantiate_err_view1' panicked at 'The window end is outside of the matrix range', src/linalg/basic/matrix.rs:61:5
stack backtrace:
```

src/linalg/basic/matrix.rs
Mec-iS: @EdmundsEcho
IMO the new methods should be added in `impl DenseMatrix` after the last currently implemented method, for example:
```
    ///  <ADD PROPER DOC HERE> 
    fn is_valid_matrix_window(
        &self,
        mrows: usize,
        mcols: usize,
        vrows: &Range<usize>,
        vcols: &Range<usize>,
    ) -> bool
```
So that they can be called from the views constructors as:
```
if m.is_valid_matrix_window(mrows, mcols, &vrows, &vcols) { ... }
```

src/linalg/basic/matrix.rs
EdmundsEcho: They pass in a release mode. The assertions can be guarded with #[cfg debug]. So, only when the debug flag is active will the assertions kick-in.

Mec-iS: tests should pass by simply running `cargo test` from the command line, if they don't we should change something or in the tests configuration or in the debug assertions to make them pass.
I would remove the debug assertions as they make the code less readable, assertions should be present only in test functions.

EdmundsEcho: I'll go ahead and update the branch so that the tests pass in non-optimized mode as well as release.

EdmundsEcho: The tests now pass in in non-optimized mode.

src/linalg/basic/matrix.rs
EdmundsEcho: I tried to add documentation using `///` but could not see in the final output of the docs.  So, I expanded on the description using `//`.  Secondly, I went through the rust-doc documentation.  It's not clear how to document the functions that are not part of a struct nor trait.  What I have seen in other libs is essentially what I have done.  Keep them local, comment for internal use only.

morenol (dismissed):

Mec-iS:
releasing a new minor for these changes is not completely justified. We need to add some more implementations to this to justify a minor release.

Mec-iS:
Also: probably, in accord with the library principles, probably it is better to panic early at initialization than returning a `Result`. smartcore is meant to be as stable as possible after compilation, using a `Result` may just move problems into runtime. So there is the option of just allow controlled `panic!(..)` instead of returning a `Fail`.

EdmundsEcho:
To which function are you referring?

I get that if a config or something early is flawed, the app should exit with useful feedback. Elsewhere, the user can "opt-in" to a panic by using "unwrap". 

Here is a link that might be useful https://doc.rust-lang.org/book/ch09-03-to-panic-or-not-to-panic.html. I see a line between if "dangerous/harmful" then panic, otherwise Err. Err is a valid answer given the input. Panic is not generally expected behavior. Furthermore "fail early" is not what we are talking about, it's more "at the time a mistake is discovered" what does the lib return. 

As a user of the library, I would expect to be able to send commands over time. A user (or service), might "get it wrong". Does that mean the user cannot "try again"?

In the event I run this in Tokio, I would not want the thread to panic (very bad and tough to navigate), but rather give me the answer - that can include: Err could not process b/c data was wrong, a match branch I can recover from. 

Runtime errors in py is the reason I moved to Rust.  *Returning Err is not a runtime error like we experience in py*. It's only a decision on what to return - app crash/panic or Err. 

More generally about runtime errors: by definition of the compile-time checks in Rust, they don't equate to py (incomparable). 

Right now the lib is fragile because it doesn't have strong typing through and through (likely mostly on the result side of the process). It will get better over time where the app does not instantiate a bad state (like what we have with DenseMatrix in the 0.4). Until then let the user know with an Err. For instance getting results is a fragile process. 

What do you think?

morenol:
I also think that a Result is better than panic for this case. Also agree that we could delay the crates.io release until we have more things that we could add to a breaking change release

EdmundsEcho:
I agree there is more that can be done for the next major release. I have some ideas for how to add to the changes for 0.4.  Much of the ideas come from what `Axum` has done for `http`.  

1. Reporting 

I tried to use the coefficients and intercept to output probabilities (not binary classes) and was surprised by how fragile the process was.  e.g., `get_row(0)`, `get((0,0))` requires a lot of "under the hood" knowledge not to mention the gymnastics required to extract values from `&X`.  

One way to go from `a -> b -> c` without the user being aware, is create what I call "transient types".  Using the transient types you build implementations of `std::convert::From`.  This way, the end user can do something like:

```rust
let coefficients: Vec<_> = lr.coefficients().into();
let intercept: f64 = lr.intercept().into();

// or 
let coefficients: Vec<_> = lr.coefficients().iter().collect();
let intercept: f64 = lr.intercept(); // casts to f64
```

There are other "best practices" for increasing the usability of each method by using traits such as `AsRef`, `Borrow` that I wonder might be worth looking through.

2. Augment what is included in a report e.g., F stat.  I'm not all familiar with the subtle differences between ML and traditional stat modeling.  I come from the latter.  The incremental effort to report on "common sense" "checks" that we should be interested in is likely good fodder for discussion.  It might be a feature, or the user choosing to "see" more of what has already been computed in the default features.

3. Move away from building your own underlying data structures for hosting and executing common math operations e.g., use ndarray.  We are using `argmin`.  We should do the same for the underlying data hosting structs.  

4. Implement "the builder" pattern, like what we experience in many Rust crates, e.g., build a http request.  I see what we are doing here in a similar vein.  This will augment the capacity to build a "default" for whatever analysis that is "more often" what you want. This would also require augmenting the documentation to fully explain the additional options (accelerated by pointing users to generically available descriptions of the concept on established stats/ml sites?).

5. All in all, it might be useful to "draw a line" where you want to be able to rely on modules that emit `Result` vs other, internal private methods that only `assert` to convey a critical dependence, but that can be made "sound" due to internal, actively monitored assertions.  In general, outside "monad", inside "pure".

For the next major after this, or the next after that :)), I would start thinking more about performance.  I hit performance issues with 50k x 100 sample size.  Part of this would include a review of when we borrow vs take ownership.  Ironically, I wonder if there is too much effort to avoid copies.  My main point is this, it's still too early to worry about performance in my view.  Just note, we have room to improve when the time comes.  Unlike python, it will actually make the code both better and more readable, not more convoluted :))

Finally, I've surveyed several packages to accomplish what I need for building stat models.  It's a fragmented space with many "half-way" efforts.  I like what you all have accomplished.  I believe the core strength is _assembly_: `data -> design -> results`.  I see `Axum` in `smartcore`.  They assemble the best while focusing on a productive, type-safe, Rust user experience.  

I hope this input is helpful.  I spent a many years working in Rust.  I'm happy to help however you all like.

Mec-iS:
Thanks for your ideas. I am glad you found `smartcore` interesting and you plan to contribute more!

I believe the library can be very useful in different domains and I am now focusing into using the library for applications to see what it can do. My idea is to keep the library different from the others on multiple dimensions. I think there is already a lot of effort into making things work for "larger" high-level use-cases and that side-tracking this approach can be very interesting in the long term for domains that are maybe not in the "big data" scenario; `polars` and `linfa` are already doing great on that, we don't want to overlap too much with their use cases; please see the [CONTRIBUTING](https://github.com/smartcorelib/smartcore/blob/development/.github/CONTRIBUTING.md). 
I am currently taking feedback from embedded developers and audio technologists mostly, focusing on real-time processing as an objective; so the use cases I have in mind are more in the context of low-level, small devices, highly portable, high-frequency computations. For example, this field of applications probably needs more something like the latest research on dimensionality reduction techniques than supporting billions of features, or else making ML algorithms more useful on tabular datasets. In this context user experience is a bit of a second-order problem but as far as resource usage stays and these levels (both in terms of time but maybe more important for small devices also _space_ complexity) or improve I am in general OK with better APIs and more features.

_It would be great to have more idiomatic Rust, I think let's start with the transient types and the builder pattern_, after that we want probably to set up good profiling (we have a bench repo but it is very limited and difficult to run, it would be nice to have something like tarpaulin for profiling, that runs with CI/CD). Vlad, that created the library, is also of the idea that we should drop our custom traits system on arrays but this is something we can see later as far as we provide the bindings to `ndarray`.
Also it would be nice to have finally fixed `predict_proba` #211, but I don't know the functionality well enough to know if it is a problem with numerical stability of the library or the implementation itself.

Welcome! @EdmundsEcho

EdmundsEcho:
To better understand the design intent was implemented, may I ask 

1. what the benefits of using a trait `Array2d` instead of a concrete type that works optimally with the solver? e.g., why not use `DenseMatrix` and convert whatever users want to this standard? 
2. are all of the methods in the `Array`-related traits used internally?  Or are they there to support pre-processing?
3. how do you see the line between pre-processing / post-processing and smartcore?
4. why not use argmin?

Mec-iS:
> To better understand the design intent was implemented, may I ask
> 
>     1. what the benefits of using a trait `Array2d` instead of a concrete type that works optimally with the solver? e.g., why not use `DenseMatrix` and convert whatever users want to this standard?
> 
>     2. are all of the methods in the `Array`-related traits used internally?  Or are they there to support pre-processing?
> 
>     3. how do you see the line between pre-processing / post-processing and smartcore?
> 
>     4. why not use argmin?

1. I think it is because to take advantage of rust zero-cost abstractions and provide zero-copy slicing and iterators, also it is possible this way to implement other custom concrete types (like a `SparseMatrix` for example)
2. They are used internally but they can also be used to define other types. I would keep the custom trait system as we may tend to have a no-dependency (or even a `no_std`) feature/fork at some point in the future
3. the library always focused on linear algebra and matrices manipulation, the rest is left to the user. Probably we can add anything that is useful behind feature configurations but the core functionalities have to stay stand-alone for minimal installation.
4. I guess it didn't exist at the time (2018), check [discussions](https://github.com/smartcorelib/smartcore/discussions) for other hypothesis of possible integrations. Again, using any library that is based on `ndarray` in the internals will make everything dependant on that, and I don't know if it's worth as we would just overlap other libraries' use cases and working patterns.

Mec-iS:
@morenol please take a look

CHANGELOG.md
```diff
@@ -4,6 +4,13 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [0.4.0] - 2023-04-05
+
+## Added
+- WARNING: Breaking changes!
+- `DenseMatrix` constructor now returns `Result` to avoid user instantiating inconsistent rows/cols count
+- Every call to `DenseMatrix::new` and `DenseMatrix::from_2d_array` needs to be unwrapped with `unwrap()`, see tests
```
morenol: I think that these two changes require to update to 0.4

morenol: I would say that instead of  saying that we need `unwrap`, we could remove line 11, and say that this two methods return a Result since `unwrap` is just one of the ways to deal with `Result`

Mec-iS: right! thanks

morenol:
I think that we need to run `cargo fmt`

morenol (approved):

PR #258, opened 2023-04-04
Fix 256: Makes instantiating DenseMatrix and others fallible

EdmundsEcho:
Fixes issue 256

Several new tests were added to the `src/linalg/basic/matrix.rs` module.
All tests pass.

Fixes #

### Checklist
- [?] Coverage and Linting have been applied 

### Current behaviour

`DenseMatrix` relies on the user to properly input the dimensions of the matrix.  In the event a mistake is made, the results will either be wrong, or generate an "index out of bounds" error.

### New expected behaviour

Update the constructors for DenseMatrix.  Set return value to `Result<DenseMatrix, Failed>` to reflect the fallible nature of the construction.  This is true because the input relies on user input to align with the data provided.  In other words, there is no way to infer and/or correct the user input when the implied matrix shape does not align with the data.

### Change logs
Clearly a breaking change.  However, this is a critically important change.  There is no way around it.  Before the change, it's possible to create references to invalid memory.  This update leverages the current error generating infrastructure. In the event the user input is nonsensical (introduces a contradiction) given the actual data, a `Failed::InputError` is returned.  

DenseMatrix, DenseMatrixView, DenseMatrixMutView, new and from_2d_array methods now validate the input and return a fallible result.
 
For future use, added a `Failed::InvalidState` to the enum.  This may be useful in the development branch to safely assert and otherwise flag when critical to "be in" a valid state.

src/linalg/basic/matrix.rs was the core change.  All of the tests, throughout the package were updated to reflect the new fallible construction.  This was prolific, but simple: tagging `.unwrap()` to the constructions used in the tests.


src/linalg/basic/matrix.rs
```diff
@@ -304,6 +384,8 @@ where
 impl<T: Debug + Display + Copy + Sized> Array<T, (usize, usize)> for DenseMatrix<T> {
     fn get(&self, pos: (usize, usize)) -> &T {
         let (row, col) = pos;
+        let idx_target = col * self.nrows + row;
```
morenol, line 387: it seems that `idx_target` is not used? maybe it was used for debugging purposes?

src/linalg/basic/matrix.rs
```diff
@@ -49,27 +51,75 @@ pub struct DenseMatrixMutView<'a, T: Debug + Display + Copy + Sized> {
...
-            )
+    fn new(
+        m: &'a DenseMatrix<T>,
+        vrows: Range<usize>,
+        vcols: Range<usize>,
+    ) -> Result<Self, Failed> {
+        let (mrows, mcols) = m.shape();
+
+        if is_valid_matrix_window(mrows, mcols, &vrows, &vcols) {
+            Err(Failed::input(&format!(
```
morenol, line 109: There are some linting issues like this across the code, could you fix it
```
useless use of `format!`
```

morenol (commented):
sounds good to me. Could you run `cargo clippy --fix` and `cargo fmt` to solve some linting issues? 

Maybe this requires to bump smartcore to 0.4.0 @Mec-iS

Mec-iS:
merged into #257

src/linalg/basic/matrix.rs
EdmundsEcho: I saw that. I kept it "as is" for now for 2 reasons:
1. I replicated how the errors were generated elsewhere  (e.g., src/decomposition/svd.rs
```rust
        if parameters.n_components >= p {
            return Err(Failed::fit(&format!(
                "Number of components, n_components should be < number of attributes ({p})"
            )));
        }
```

2. using `format` is preferred when it includes a variable (~"string literal") to help inform the error.  I updated the message to echo the matrix shape.

PR #263, opened 2023-04-24
chore: fix clippy warnings from Rust release 1.69

morenol:
Fixed running `cargo clippy --fix`


morenol:
@Mec-iS Please review and merge if you think that it is ok

Mec-iS (approved):

PR #283, opened 2024-11-19
chore: fix clippy

morenol:


Mec-iS (dismissed):

Mec-iS:
@morenol what do we do with these doc-empty errors?

morenol:
> @morenol what do we do with these doc-empty errors?

I have trying to fix them but they are a lot and I am not familiar with all of them

PR #288, opened 2025-01-20
Implement predict_proba for RandomForestClassifier

### LINKED ISSUE #50: Implement a new method that predicts probabilities, where it makes sense

Mec-iS:
Fixes #50
implement `predict_proba` for `RandomForestClassifier`

### Checklist


Mec-iS:
@morenol could you take a look to the new tests? I get inconsistent errors from some platforms

PR #299, opened 2025-04-24
Update Cargo.toml

morenol:


Mec-iS (dismissed):

Mec-iS:
the new checkout operation is failing?

morenol:
> the new checkout operation is failing?

it looks like `cache@v2` has issues, I update that to v4 and also updated checkout action

PR #305, opened 2025-06-03
Made SVM multiclass

DanielLacina:
I split the SVM classification into multiple binary classifications using the one for all strategy.

PR #306, opened 2025-06-03
SVC multiclass

DanielLacina:
Implemented the multiclass feature for SVC using a one for all approach.


DanielLacina:
Had to close the previous pr because I sent it on the wrong branch.

Mec-iS:
https://github.com/smartcorelib/smartcore/pull/305#issuecomment-2937766392

DanielLacina:
The issue with Array2 is that the type within the Array is inferred and static. If a Vec<u 32> is passed in, then the application won't be able to transform the labels to: {1, -1} in order for them to be used for binary classification. Originally, you had an assert statement to validate the data passed in had labels: -1 and 1, but now the SVC, with it being multiclass, can accept a wide variety of labels.

Mec-iS:
> The issue with Array2 is that the type within the Array is inferred and static. If a Vec<u 32> is passed in, then the application won't be able to transform the labels to: {1, -1} in order for them to be used for binary classification. 

Array2 is an abstraction for a 2D vector, it can be used for any instance supported by Vec, it just need to be implemented.

Test Driven Development (TDD) should be followed. Every time you implement something new you have to be sure to support existing behaviour. If there is no test for the operations you are changing, you should add it. Also please add one or more tests when you implement something new.

When I try to run the tests in the module I get:
```
running 3 tests
test svm::svc::tests::svc_fit_predict ... ok
test svm::svc::tests::svc_fit_predict_rbf ... ok
test svm::svc::tests::svc_fit_decision_function ... FAILED

successes:

successes:
    svm::svc::tests::svc_fit_predict
    svm::svc::tests::svc_fit_predict_rbf

failures:

---- svm::svc::tests::svc_fit_decision_function stdout ----
thread 'svm::svc::tests::svc_fit_decision_function' panicked at src/svm/svc.rs:1033:9:
assertion failed: y_hat[1] < y_hat[2]
```


This is a *reference implementation* as generated by my LLM, you can start from this as it is not fully implemented (ie. it is an example and needs to be implemented using generic types like TX and TY). Please note that you should not modify the existing struct but instead create new structs to handle the multiclass possibility. This implementation suggests to use a 1D Vec but you should check if it correct (if it is right there is no need to use a 2D Vec):

---

To implement a multiclass Support Vector Classification (SVC) in Rust using smartcore, we can adopt the **one-vs-one (OvO)** strategy, which trains binary classifiers for each pair of classes. Here's a complete implementation:

```rust
use smartcore::svm::svc::{SVC, SVCParameters};
use smartcore::linalg::{BaseVector, Matrix, MatrixTrait};
use smartcore::metrics::accuracy;
use smartcore::dataset::iris::load_dataset;

// Multiclass SVC using One-vs-One strategy
struct MulticlassSVC {
    classifiers: Vec<SVC<f64, DenseMatrix<f64>, Vec<f64>>>,
    classes: Vec<u32>,
}

impl MulticlassSVC {
    pub fn fit(
        x: &DenseMatrix<f64>,
        y: &Vec<u32>,
        params: &SVCParameters<f64>,
    ) -> Result<Self, Failed> {
        let classes = y.iter().unique().sorted().collect::<Vec<_>>();
        let mut classifiers = Vec::new();

        // Generate all class pairs
        for (i, &class1) in classes.iter().enumerate() {
            for &class2 in classes.iter().skip(i + 1) {
                // Filter samples for current class pair
                let indices: Vec<usize> = y.iter()
                    .enumerate()
                    .filter(|(_, &c)| c == class1 || c == class2)
                    .map(|(i, _)| i)
                    .collect();

                let x_filtered = x.select_rows(&indices);
                let y_filtered: Vec<f64> = indices.iter()
                    .map(|&i| if y[i] == class1 { 1.0 } else { -1.0 })
                    .collect();

                // Train binary classifier
                let mut clf = SVC::fit(&x_filtered, &y_filtered, params.clone())?;
                classifiers.push((class1, class2, clf));
            }
        }

        Ok(Self { classifiers, classes })
    }

    pub fn predict(&self, x: &DenseMatrix<f64>) -> Vec<u32> {
        let mut votes = vec![HashMap::new(); x.shape().0];
        
        for (class1, class2, clf) in &self.classifiers {
            let preds = clf.predict(x).unwrap();
            
            for (i, &p) in preds.iter().enumerate() {
                let vote = if p > 0.0 { *class1 } else { *class2 };
                *votes[i].entry(vote).or_insert(0) += 1;
            }
        }

        votes.iter()
            .map(|v| *v.iter().max_by_key(|(_, &count)| count).unwrap().0)
            .collect()
    }
}

// Example usage with Iris dataset
fn main() -> Result<(), Failed> {
    let iris = load_dataset();
    let (x_train, x_test, y_train, y_test) = train_test_split(
        &iris.data,
        &iris.target,
        0.8,
        true,
        Some(42),
    );

    let params = SVCParameters::default()
        .with_c(200.0)
        .with_kernel(smartcore::svm::Kernel::linear());

    let clf = MulticlassSVC::fit(&x_train, &y_train, &params)?;
    let preds = clf.predict(&x_test);
    
    println!("Accuracy: {}", accuracy(&y_test, &preds));
    Ok(())
}
```

**Key implementation details:**

1. **OvO Strategy:**
    - Creates `n_classes * (n_classes - 1) / 2` binary classifiers[^2][^3]
    - Uses filtered subsets of data for each class pair[^6]
2. **smartcore Integration:**
    - Uses `SVC` with configurable parameters (C, kernel)[^5]
    - Handles `DenseMatrix` input as per smartcore's data requirements[^1]
3. **Prediction Aggregation:**
    - Implements voting system across all binary classifiers[^3][^6]
    - Handles ties by selecting first majority class

**Advantages over naive implementation:**

- Maintains smartcore's API conventions[^1]
- Avoids unsafe code and complex lifetimes
- Compatible with smartcore's linear algebra traits
- Supports all kernel types available in smartcore[^5]

This implementation follows smartcore's design principles by:

- Using pure Rust without external dependencies[^1]
- Maintaining a Pythonic/scikit-learn-like API[^1]
- Supporting TDD through clear input/output contracts
- Working efficiently with smartcore's matrix types[^1]

For production use, you'd want to add:

- Model persistence (serialization/deserialization)
- Class weighting support
- Parallel training of binary classifiers (using rayon)
- More sophisticated tie-breaking strategies


[^2]: https://scikit-learn.org/stable/modules/svm.html

[^3]: https://www.nb-data.com/p/one-vs-all-vs-one-vs-one-which-multi

[^4]: https://stackoverflow.com/questions/62805976/how-to-perform-multi-class-svm-in-python

[^5]: http://smartcorelib.org/user_guide/supervised.html

[^6]: https://www.baeldung.com/cs/svm-multiclass-classification

[^7]: https://github.com/smartcorelib/smartcore/milestone/1

[^8]: https://github.com/smartcorelib/smartcore/issues

[^9]: https://users.rust-lang.org/t/using-rust-for-ai/69340

[^10]: https://stackoverflow.com/questions/59497115/svc-multiclass-classification-ovo-decision-function-explanation

[^11]: https://stackoverflow.com/questions/35076586/when-should-one-use-linearsvc-or-svc

DanielLacina:
Appreciate the guide.

DanielLacina:
It doesn't look like the DenseMatrix has the select_rows method.

DanielLacina:
I assume you want the multiclass struct to take in any generic Array2 object as x.

Mec-iS:
> It doesn't look like the DenseMatrix has the select_rows method.

it has the `get_row` method, see `src/linalg/basic/matrix.rs`. Please read the API for `DenseMatrix`.

> I assume you want the multiclass struct to take in any generic Array2 object as x.

it depends what you are going to do with the `y` parameter, in this case using `&Vec` it is OK I guess. Just try and see.

DanielLacina:
Alright. That helps a ton.

DanielLacina:
I have to clone the matrix in order to create a version of the matrix that filters out the rows that don't have one of the two labels. The issue is that it creates a local copy of the matrix, but the SVC struct only accepts references. This leads to the lifetime of the matrix being shorter than the SVC which is not allowed.

DanielLacina:
I think I found a solution.

src/svm/svc.rs
```diff
@@ -84,12 +85,107 @@ use serde::{Deserialize, Serialize};
...
+use crate::linalg::basic::arrays::ArrayView1;
+use crate::linalg::basic::arrays::{Array, Array1, Array2, MutArray};
+use crate::linalg::basic::matrix::DenseMatrix;
 use crate::numbers::basenum::Number;
 use crate::numbers::realnum::RealNumber;
 use crate::rand_custom::get_rng_impl;
 use crate::svm::Kernel;
 
+pub struct MultiClassSVCParameters<'a> {
+    xs_filtered: Vec<DenseMatrix<f64>>,
```
Mec-iS, line 97: this needs to be implemented using the generic `TX` `TY`. It should work for all supported types, not just f64.

src/svm/svc.rs
```diff
@@ -84,12 +85,107 @@ use serde::{Deserialize, Serialize};
...
+            let x_filtered = params.xs_filtered.get(i).unwrap();
+            let svc = SVC::fit(x_filtered, y_filtered, params.svc_params);
+        }
+        Ok(Self {
+            parameters: &params,
+            classifiers,
+        })
+    }
+
+    pub fn predict(&self, x: &DenseMatrix<f64>) -> Vec<i64> {
```
Mec-iS: this should be implemented for the generic types `TX` `TY`, not just for f64. See the other implementations for references.

src/svm/svc.rs
```diff
@@ -84,12 +85,107 @@ use serde::{Deserialize, Serialize};
...
+                }
+            }
+        }
+        polls
+            .iter()
+            .map(|v| *v.iter().max_by_key(|(_, count)| **count).unwrap().0)
+            .collect()
+    }
+}
+
```
Mec-iS, line 277: please provide multiple tests using the provided dataset or simple examples.
Also consider writing a bench and adding it to the `smartcore-benches` repository

src/svm/svc.rs
```diff
@@ -84,12 +85,107 @@ use serde::{Deserialize, Serialize};
...
-use crate::linalg::basic::arrays::{Array1, Array2, MutArray};
+use crate::linalg::basic::arrays::ArrayView1;
+use crate::linalg::basic::arrays::{Array, Array1, Array2, MutArray};
+use crate::linalg::basic::matrix::DenseMatrix;
 use crate::numbers::basenum::Number;
 use crate::numbers::realnum::RealNumber;
 use crate::rand_custom::get_rng_impl;
 use crate::svm::Kernel;
 
+pub struct MultiClassSVCParameters<'a> {
```
Mec-iS: good work! much better 👍🏼

DanielLacina:
The code doesn't work btw. Im facing an issue where the shapes of whatever vector type objects aren't the same. I did learn a lot by playing around with the code. Ig I can say Im one step ahead.

Mec-iS:
good progress 👍 keep going

DanielLacina:
I noticed there's a lot of parts of the code where you loop through each row of X and convert them to some type. I think that could be the reason why the SVC is so slow.

DanielLacina:
I know it's not relevant to this PR but it's really important.

DanielLacina:
Where in the repo would I store my dataset-related tests?

DanielLacina:
I wrote one unit test that tests the whole process.

DanielLacina:
Do you want the benchmark to take in the Iris dataset as input? Sorry, I haven't benchmarked before.

Mec-iS:
> Where in the repo would I store my dataset-related tests?

tests go at the bottom of the module in the `mod tests {...}` submodule, as suggested by Rust best practices and implemented in the rest of the library

Mec-iS:
> Do you want the benchmark to take in the Iris dataset as input? Sorry, I haven't benchmarked before.

you can follow the current practice in the `smartcore-benches` repository where the other benches are. Please open a PR there for details.

Mec-iS:
just ask the LLM to write tests for this module and see what it returns; it is usually good with basic usages.

DanielLacina:
I sent a pr for the smartcore-benches repo.

DanielLacina:
> > Where in the repo would I store my dataset-related tests?
> 
> tests go at the bottom of the module in the `mod tests {...}` submodule, as suggested by Rust best practices and implemented in the rest of the library

Aren't those for unit tests? What about integration tests? I assume any test involving a dataset would be an integration test.

DanielLacina:
PR is ready fo review.

Mec-iS:
> Aren't those for unit tests? What about integration tests? I assume any test involving a dataset would be an integration test.

i don't know if you can technically call those you mentioned integration tests. btw the ones involving datasets need to be flagged with the `datasets` feature, see other modules and cargo.toml

DanielLacina:
I created a unit test so everything is good

Mec-iS:
Thanks for your contribution. The code is returning the correct results as compared to `scikit-learn`. Nice one :+1:

As specified in `CONTRIBUTING.md` please run `rustfmt src/*.rs` and `cargo clippy --all-features -- -Drust-2018-idioms -Dwarnings` otherwise  the CI/CD tests are failing.

Also please double check that you have checked and possibly reused the traits defined in base library:
```
#### linalg/traits
The traits in `src/linalg/traits` are closely linked to Linear Algebra's theoretical framework. These traits are used to specify characteristics and constraints for types accepted by various algorithms. For example these allow to define if a matrix is `QRDecomposable` and/or `SVDDecomposable`. See docstring for referencese to theoretical framework.

As above these are all traits and by definition they do not allow instantiation. They are mostly used to provide constraints for implementations. For example, the implementation for Linear Regression requires the input data `X` to be in `smartcore`'s trait system `Array2<FloatNumber> + QRDecomposable<TX> + SVDDecomposable<TX>`, a 2-D matrix that is both QR and SVD decomposable; that is what the provided strucure `linalg::arrays::matrix::DenseMatrix` happens to be: `impl<T: FloatNumber> QRDecomposable<T> for DenseMatrix<T> {};impl<T: FloatNumber> SVDDecomposable<T> for DenseMatrix<T> {}`. 
```

src/svm/svc.rs
```diff
@@ -84,12 +85,208 @@ use serde::{Deserialize, Serialize};
...
+    /// classifiers to determine the final prediction for each data point.
+    ///
+    /// # Arguments
+    /// * `x` - A reference to the input features (2D array) for which to make predictions.
+    ///
+    /// # Returns
+    /// A `Result` containing a `Vec` of predicted class labels (`TX`) or a `Failed` error.
+    ///
+    /// # Panics
+    /// Panics if the model has not been fitted (`self.classifiers` is `None`).
```
Mec-iS: this should not panic but return an error (see the error harness in `smartcore/src/error/mod.rs`). Panics should be limited to unknowns so that we know that something non-expected happened. Everything that is expected should return an error.

src/svm/svc.rs
```diff
@@ -226,7 +425,7 @@ impl<'a, TX: Number + RealNumber, TY: Number + Ord, X: Array2<TX>, Y: Array1<TY>
         y: &'a Y,
         parameters: &'a SVCParameters<TX, TY, X, Y>,
     ) -> Result<Self, Failed> {
-        SVC::fit(x, y, parameters)
+        SVC::fit(x, y, parameters, None)
```
Mec-iS, line 416: why the signature for `SVC::fit` changed? this breaks the API and breaks all code that is already developed for previous versions. Never change the existing API, this is a breaking change that requires a major release (we had a major release with v0.4.0 when we changed the API to the new stable one).

The signature for `SVC::fit` is:
```
fn fit(
        x: &'a X,
        y: &'a Y,
        parameters: &'a SVCParameters<TX, TY, X, Y>,
    ) 
  ```
  and such shall remain or all the previous applications written with the library will break and cause a lot of problems to all the users already using SVC.
Please be attentive in keeping the existing API, one of the main point of smartcore is stability. 
In this case it is better to have some code duplication than breaking the API.

File: src/svm/svc.rs [hunk 7]
```diff
@@ -265,27 +465,22 @@ impl<'a, TX: Number + RealNumber, TY: Number + Ord, X: Array2<TX> + 'a, Y: Array
...
-                    FailedError::ParametersError,
-                    "Class labels must be 1 or -1",
-                ));
+        let (indices, classes) = if let Some(multiclass_config) = multiclass_config {
+            let classes = multiclass_config.classes;
+            (Some(multiclass_config.indices), classes)
+        } else {
+            let classes = y.unique();
+            if classes.len() != 2 {
+                return Err(Failed::fit(&format!(
```
Mec-iS, line 462: good. this is good error reporting.

src/svm/svc.rs
DanielLacina:
```rust
impl<'a, TX: Number + RealNumber, TY: Number + Ord, X: Array2<TX>, Y: Array1<TY>>
    SupervisedEstimatorBorrow<'a, X, Y, SVCParameters<TX, TY, X, Y>> for SVC<'a, TX, TY, X, Y>
{
    fn new() -> Self {
        Self {
            classes: Option::None,
            instances: Option::None,
            parameters: Option::None,
            w: Option::None,
            b: Option::None,
            phantomdata: PhantomData,
        }
    }
    fn fit(
        x: &'a X,
        y: &'a Y,
        parameters: &'a SVCParameters<TX, TY, X, Y>,
    ) -> Result<Self, Failed> {
        SVC::fit(x, y, parameters)
    }
}```

Comment (DanielLacina) [reply src/svm/svc.rs:416]: Isnt this the API?

Comment (Mec-iS) [reply src/svm/svc.rs:416]: Your code work well and you are in the good path for a good implementation.

Although, the original API for SVC is [this one](https://github.com/smartcorelib/smartcore/blob/44424807a040681b917891b9efabb3f28fe562ef/src/svm/svc.rs#L248)

The PR changes it to:
```rust
impl<'a, TX: Number + RealNumber, TY: Number + Ord, X: Array2<TX> + 'a, Y: Array1<TY> + 'a>
    SVC<'a, TX, TY, X, Y>
{
    /// Fits SVC to your data.
    /// * `x` - _NxM_ matrix with _N_ observations and _M_ features in each observation.
    /// * `y` - class labels
    /// * `parameters` - optional parameters, use `Default::default()` to set parameters to default values.
    pub fn fit(
        x: &'a X,
        y: &'a Y,
        parameters: &'a SVCParameters<TX, TY, X, Y>,
        multiclass_config: Option<MultiClassConfig<TY>>,
    )
```
in your code `multiclass_config` is not part of the original `SVC` type, so it will break all existing code.
My idea was to keep the current `SVC` implementation unchanged so that we don't break existing code and to add a `SVCMultiClass` that implements the multiclasses classifier without touching the existing `SVC`.

So you should change your implementation for the new `SVCMultiClass` without changing anything in `SVC`.

You can try asking an LLM: "how do I change this code in a way that it doesn't change the existing SVC code"

DanielLacina: check the new pr

DanielLacina: I think I fixed everything.

PR #311, opened 2025-06-25
Hierarchal clustering

DanielLacina:


Mec-iS:
thank you! I will take a look at this asap.

Would be nice to have some way to display the dendrogram also, maybe using ASCII art of something like it.

PR #312, opened 2025-06-25
implemented hierarchal clustering

DanielLacina:


DanielLacina:
Whenever I create a certain n amount of clusters, the labels returned by this implementation and sklearn start to deviate. Is it because of the difference in numeric precision with the Python ml libraries?

DanielLacina:
My Python implementation works perfectly.

Mec-iS:
I will take a look. For now you can compare with this implementation https://github.com/rom1mouret/catclustering

src/cluster/mod.rs
```diff
@@ -4,5 +4,6 @@
 //! are more similar to other data points in the same group than those in other groups. In simple words, the aim is to segregate groups with similar traits and assign them into clusters.
 
 pub mod dbscan;
+pub mod hierarchal;
```
Mec-iS, line 7: typo: hierarchical

also better to use `agglomerative`

Mec-iS:
@DanielLacina as a pairing algorithm please use fastpair as it is implemented in the pairing module

DanielLacina:
thanks

DanielLacina:
Do you think we should add a linkage attribute to the FastPair struct?

PR #314, opened 2025-07-08
implemented xgdboost_regression

DanielLacina:


src/xgboost/xgb_regressor.rs
```diff
@@ -0,0 +1,756 @@
...
+    /// * `data` - The full training dataset.
+    /// * `g` - Gradients for all samples.
+    /// * `h` - Hessians for all samples.
+    /// * `idxs` - The indices of the samples belonging to the current node.
+    /// * `max_depth` - The maximum remaining depth for this branch.
+    /// * `min_child_weight` - The minimum sum of hessians required in a child node.
+    /// * `lambda` - L2 regularization term on weights.
+    /// * `gamma` - Minimum loss reduction required to make a further partition.
+    pub fn fit(
+        data: &Vec<Vec<f64>>,
```
Mec-iS: why are we using allocation here?

all the `fit(...)` methods should have a similar signature and use the `Array` types. Like you did in:

```rust
impl<TX: Number, TY: Number, X: Array2<TX>, Y: Array1<TY>> AgglomerativeClustering<TX, TY, X, Y> {
    /// Fits the agglomerative clustering model to the data.
    ///
    /// # Arguments
    /// * `data` - A reference to the input data matrix.
    /// * `parameters` - The parameters for the clustering algorithm, including `n_clusters`.
    ///
    /// # Returns
    /// A `Result` containing the fitted model with cluster labels, or an error if
    pub fn fit(data: &X, parameters: AgglomerativeClusteringParameters) -> Result<Self, Failed> {
        let (num_samples, _) = data.shape();
        let n_clusters = parameters.n_clusters;
        if n_clusters > num_samples {
            return Err(Failed::because(
                FailedError::ParametersError,
                &format!(
                    "n_clusters: {n_clusters} cannot be greater than n_samples: {num_samples}"
                ),
            ));
        }
```

Same in other parts, whenever we use Vec we should ask if it is not the case to use `Array`. Vec can be used for labels and sequence that we are sure are relatively small order of magnitude (usually 100-300 elements). For everything that is 2D or require large arrays, the `Array` interface should be implemented using `DenseMatrix` or the `Array` traits.

src/xgboost/xgb_regressor.rs
```diff
@@ -0,0 +1,756 @@
...
+                *best_split_score = gain;
+                *best_threshold = (x_i + x_i_next) / 2.0;
+                *best_feature_idx = feature_idx;
+            }
+        }
+    }
+
+    /// Predicts the output values for a dataset.
+    pub fn predict(&self, data: &[Vec<f64>]) -> Vec<f64> {
+        data.iter().map(|row| self.predict_for_row(row)).collect()
```
Mec-iS: as above this iteration should use the `.iterator(0).copied()` implementation. There is a custom iterator for Array, but if you don't use Array you cannot access it. This is also important for compatibility with other libraries like `ndarray`

src/xgboost/xgb_regressor.rs
```diff
@@ -0,0 +1,756 @@
...
+    parameters: Option<XGRegressorParameters>,
+    _phantom_ty: PhantomData<TY>,
+    _phantom_tx: PhantomData<TX>,
+    _phantom_y: PhantomData<Y>,
+    _phantom_x: PhantomData<X>,
+}
+
+impl<TX: Number + PartialOrd, TY: Number, X: Array2<TX>, Y: Array1<TY>> XGRegressor<TX, TY, X, Y> {
+    /// Fits the XGBoost model to the training data.
+    pub fn fit(data: &X, y: &Y, parameters: XGRegressorParameters) -> Result<Self, Failed> {
```
Mec-iS, line 508: exactly. every `fit` method should have a signature that looks like this, `X` is an `Array2<TX>` trait type.

Every other method that iterates over data should use the `Array` interface.

Mec-iS:
Thanks 👍🏼

DanielLacina:
resolved the problems

DanielLacina:
I think the lints are resolved

Mec-iS (approved):
Wondeful!

PR #315, opened 2025-07-09
Extra trees

DanielLacina:


DanielLacina:
I think I'm going to split the decision tree refactoring and extra trees implementation into two different pull requests.

PR #316, opened 2025-07-09
refactored decision tree into reusable components

DanielLacina:


src/tree/base_tree_regressor.rs
```diff
@@ -0,0 +1,551 @@
...
+    }
+    /// Get estimate of intercept, return value
+    fn depth(&self) -> u16 {
+        self.depth
+    }
+}
+
+#[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
+#[derive(Debug, Clone)]
+struct Node {
```
Mec-iS, line 78: this should problably be `Node<TX>`, why it should be limited to f64? in general everytime you write a fixed numerical type you should ask yourself if it is generalisable.

src/tree/base_tree_regressor.rs
```diff
@@ -0,0 +1,551 @@
...
+            split_score: Option::None,
+            true_child: Option::None,
+            false_child: Option::None,
+        }
+    }
+}
+
+impl PartialEq for Node {
+    fn eq(&self, other: &Self) -> bool {
+        (self.output - other.output).abs() < f64::EPSILON
```
Mec-iS, line 102: again, this will work only for f64?

src/tree/base_tree_regressor.rs
```diff
@@ -0,0 +1,551 @@
...
+use serde::{Deserialize, Serialize};
+
+use crate::error::Failed;
+use crate::linalg::basic::arrays::{Array1, Array2, MutArrayView1};
+use crate::numbers::basenum::Number;
+use crate::rand_custom::get_rng_impl;
+
+#[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
+#[derive(Debug, Clone, Default)]
+pub enum Splitter {
```
Mec-iS, line 19: missing docstring

src/tree/base_tree_regressor.rs
```diff
@@ -0,0 +1,551 @@
...
+    Random,
+    #[default]
+    Best,
+}
+
+#[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
+#[derive(Debug, Clone)]
+/// Parameters of Regression base_tree
```
Mec-iS, line 27: this should explain a little more like:
```
/// Parameters for configuring the behavior of a decision tree regressor.
/// 
/// This struct controls various aspects of tree construction including
/// depth limits, sample requirements, and splitting strategies.
```

src/tree/base_tree_regressor.rs
```diff
@@ -0,0 +1,551 @@
...
+            n += *sample_i;
+            sum += *sample_i as f64 * y_m.get(i).to_f64().unwrap();
+        }
+
+        let root = Node::new(sum / (n as f64));
+        nodes.push(root);
+        let mut order: Vec<Vec<usize>> = Vec::new();
+
+        for i in 0..num_attributes {
+            let mut col_i: Vec<TX> = x.get_col(i).iterator(0).copied().collect();
```
Mec-iS, line 218: this one could be just maybe?

```
let mut col_view: Vec<f64> = x.get_col(i).iterator(0).copied().collect();
let indices = col_view.argsort_mut();
```

src/tree/base_tree_regressor.rs
```diff
@@ -0,0 +1,551 @@
...
+{
+    /// Get nodes, return a shared reference
+    fn nodes(&self) -> &Vec<Node> {
+        self.nodes.as_ref()
+    }
+    /// Get parameters, return a shared reference
+    fn parameters(&self) -> &BaseTreeRegressorParameters {
+        self.parameters.as_ref().unwrap()
+    }
+    /// Get estimate of intercept, return value
```
Mec-iS, line 70: I don't think this comment is correct, please double-check

Mec-iS:
Thank you!

This is a nice starting point so to make possible to implement:
* Classification Trees
* Ensemble Methods: random forests
* Feature Selection: `mtry` can be used for feature sampling

Mec-iS:
This opens up to implementations like:
```rust
trait TreeRegressor<TX, TY, X, Y> {
    fn fit_tree(&self, x: &X, y: &Y, samples: Vec<usize>) -> Result<Self, Failed>;
    fn predict_tree(&self, x: &X) -> Result<Y, Failed>;
}

trait EnsembleRegressor<TX, TY, X, Y> {
    fn fit_ensemble(&self, x: &X, y: &Y, n_estimators: usize) -> Result<Self, Failed>;
    fn predict_ensemble(&self, x: &X) -> Result<Y, Failed>;
}
```
is that correct?

Recommended Implementation Priority
* Random Forest Regressor: Highest priority due to existing foundation
* Gradient Boosting Regressor: High impact for predictive performance
* Extra Trees Regressor: Low implementation cost with good diversity benefits
* Quantile Regression Trees: Valuable for uncertainty quantification
* Pruned Regression Trees: Important for model interpretability

src/tree/base_tree_regressor.rs
DanielLacina: All this code is copied and pasted from the DecisionTree file with the difference being that I use the name BaseTree instead of DecisionTree. This implementation is already in your codebase: https://github.com/smartcorelib/smartcore/blob/development/src/tree/decision_tree_regressor.rs#L129.

DanielLacina: Plus Node is an implementation detail (no pub keyword) so it doesn't matter what numeric type it contains.

src/tree/base_tree_regressor.rs
DanielLacina: Node is an implementation detail

src/tree/base_tree_regressor.rs
DanielLacina: I copied and pasted this code from the decision tree file. That would probably be more readable. Ima be honest, it was pretty hard reading the code of the decision tree implementation, but it is indeed optimal.

src/tree/base_tree_regressor.rs
DanielLacina: That's already written in the codebase.

DanielLacina: https://github.com/smartcorelib/smartcore/blob/development/src/tree/decision_tree_regressor.rs#L121

DanielLacina:
I think we should build Base models that other more complex models can build on top of.

DanielLacina:
Idk if traits are necessary. I think all we need is for example a complex model that stores the base model as an attribute. And then we can add extra logic on top of the base model.

Mec-iS (approved):

PR #321, opened 2025-07-13
Matrix allocation optimization

DanielLacina:


DanielLacina:
This is not finished but I in order to use the MatrixPool, the Dense Matrix could only accept types that have static lifetimes.

DanielLacina:
There's some situations where if you pass in borrowed values , the static lifetime condition is not satisifed.

Mec-iS:
i don't think this is the right way to go.
Let's think about this for some time to have a better understanding

DanielLacina:
Ok

Mec-iS:
or at least we need a multi-step roadmap to gradually implement this.

PR #338, opened 2025-11-22
fix precision and recall calculations

### LINKED ISSUE #323: Bug: Precision and recall calculations
**Description:**
I discovered that the Precision and Recall have their logic reversed, the Precision metric is actually calculating Recall, and the Recall metric is calculating Precision.

**The Issue:**
In the Precision implementation, when counting false positives in binary classification:

```
if y_pred.get(i) == y_true.get(i) {
    if classes == 2 {
        if *y_true.get(i) == T::one() {
            tp += 1;
        }
    } else {
        tp += 1;
    }
} else if classes == 2 {
    if *y_true.get(i) == T::one() {
        fp += 1;
    }
}
```

This checks if y_true == 1 when predictions don't match. However, this is counting false negatives (when true=1 but pred=0), not false positives. False positives should be when we predict 1 but true label is 0.

Similarly, in the Recall implementation, when counting false negatives:
```

if y_pred.get(i) == y_true.get(i) {
    if classes == 2 {
        if *y_true.get(i) == T::one() {
            tp += 1;
        }
    } else {
        tp += 1;
    }
} else if classes == 2 {
    if *y_true.get(i) != T::one() {
        fne += 1;
    }
}
```

This checks if y_true == 0 when predictions don't match, which actually counts false positives (when true=0 but pred=1), not false negatives.

**Impact:**

- Precision scores are actually Recall scores
- Recall scores are actually Precision scores
- F-beta scores with beta ≠ 1 are incorrect

cmmartin:
Fixes #323

### Checklist

### Current behaviour
- **recall**
  - binary case calculates precision instead of recall due to incorrect false negative counting
  - multiclass case computes accuracy (total correct / total) instead of macro-averaged recall
- **precision**
  - binary case calculates recall instead of precision due to incorrect false positive counting
  - multiclass case computes accuracy instead of macro-averaged precision

### New expected behaviour
- **recall**
  - binary case correctly calculates recall
  - multiclass case computes macro-averaged recall instead of accuracy
  - handles zero-division edge case
- **precision**
  - binary case correctly calculates precision
  - multiclass case computes macro-averaged precision instead of accuracy
  - handles zero-division edge case

### DIFF
--- a/src/metrics/precision.rs
+++ b/src/metrics/precision.rs
@@ -4,7 +4,9 @@
 //!
 //! \\[precision = \frac{tp}{tp + fp}\\]
 //!
-//! where tp (true positive) - correct result, fp (false positive) - unexpected result
+//! where tp (true positive) - correct result, fp (false positive) - unexpected result.
+//! For binary classification, this is precision for the positive class (assumed to be 1.0).
+//! For multiclass, this is macro-averaged precision (average of per-class precisions).
 //!
 //! Example:
 //!
@@ -19,7 +21,8 @@
 //!
 //! <script src="https://polyfill.io/v3/polyfill.min.js?features=es6"></script>
 //! <script id="MathJax-script" async src="https://cdn.jsdelivr.net/npm/mathjax@3/es5/tex-mml-chtml.js"></script>
-use std::collections::HashSet;
+
+use std::collections::{HashMap, HashSet};
 use std::marker::PhantomData;
 
 #[cfg(feature = "serde")]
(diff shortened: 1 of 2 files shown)


Mec-iS:
thanks a lot. I will review this on monday

src/metrics/precision.rs
```diff
@@ -114,7 +147,7 @@ mod tests {
         let y_pred: Vec<f64> = vec![0., 0., 1., 1., 1., 1.];
 
         let score3: f64 = Precision::new().get_score(&y_true, &y_pred);
-        assert!((score3 - 0.6666666666).abs() < 1e-8);
+        assert!((score3 - 0.5).abs() < 1e-8);
```
Mec-iS, line 150: it would be nice to take the opportunity and make these statements to use `approx` where possible

Mec-iS (approved):
💯 
Hope you will stick around!

Mec-iS:
I will merge this and create a new release

Issue #342, opened 2025-11-24
LASSO Implementation: Condition bug, missing normalization documentation, and intercept fitting option


georeth:
## Bug: Incorrect condition for dimension check
https://github.com/smartcorelib/smartcore/blob/0caa8306ff448abbbd1bae7a4fa0dae7af08afcb/src/linear/lasso.rs#L249-L253

The condition `n <= p` is inconsistent with the error message, it should be `n < p`. (my case is full-rank)

## L1 penalty normalization not documented
https://github.com/smartcorelib/smartcore/blob/0caa8306ff448abbbd1bae7a4fa0dae7af08afcb/src/linear/lasso.rs#L12

the penalty in docs is `alpha * sum(abs(beta))`, but I think it's `alpha * n * sum(abs(beta))` in the code, where `n` is the number of observations.

## Feature Request: Option to disable intercept fitting
https://github.com/smartcorelib/smartcore/blob/0caa8306ff448abbbd1bae7a4fa0dae7af08afcb/src/linear/lasso_optimizer.rs#L64

Currently, the mean is always substracted from `y`. Is it possible to add a parameter (e.g., fit_intercept: bool, or more specific substract_mean_from_y: bool) to allow users to disable intercept fitting? (force the `beta_0` parameter in the formula to 0)

Mec-iS:
thank you for reporting this issue.

georeth:
sklearn lasso uses `(1 / (2 * n_samples)) * ||y - Xw||^2_2 + alpha * ||w||_1` as loss function, and has a `fit_intercept` option. smartcore uses `(1 / n_samples) * ||y - Xw||^2_2 + alpha * ||w||_1`.

I need to solve a not well-conditioned, full-rank LASSO without intercept term. sklearn's coordinate descent failed to converge for my problem. I am trying to make smartcore solve LASSO without intercept term, but not familiar with interior point method.

Mec-iS:
As you are the one using it, you are probably in the best position to fix this.
If you have any question regarding `smartcore` internals and how to use them for this particular problem, I would be glad to answer.

georeth:
I have raised a PR for the first two points.
For the third one, I verified by experiments that it converges the desired result if I comment out the code to normalize y. I am not sure whether other changes should be made to the interior point algorithm. (in other words, I doubt if the algorithm depends on the fact that mean of y is zero.)
I will raise this in a PR after the first PR is accepted.

Mec-iS:
[...]

PR #343, opened 2025-11-26
Fix LASSO (first two of #342)

georeth:
### Checklist

### Current behaviour
Lasso::fit(X, y) with square X fails.

### New expected behaviour
allow `n==p` case

### Change logs

LASSO
* change loss function in doc to match code
* allow `n == p` case

### DIFF
--- a/src/linear/lasso.rs
+++ b/src/linear/lasso.rs
@@ -369,6 +369,7 @@ impl<TX: FloatNumber + RealNumber, TY: Number, X: Array2<TX>, Y: Array1<TY>> Las
 #[cfg(test)]
 mod tests {
     use super::*;
+    use crate::linalg::basic::arrays::Array;
     use crate::linalg::basic::matrix::DenseMatrix;
     use crate::metrics::mean_absolute_error;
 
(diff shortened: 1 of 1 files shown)


Mec-iS (commented):
please add a test specific for the behaviour as you have fixed it.

georeth:
add a test case for `n == p` case.

please fix this line:
```rust
let w = Lasso::fit(&x, &y, param).unwrap().coefficients.unwrap().iter().copied().collect_vec(); // FIXME: better way?
```
The fit result is a column vector stored in a 2d array. I didn't figure out how to get a `Vec` from a `DenseMatrix` with only one column. `get_col()` returns `ArrayView1` and `take_col()` returns `DenseMatrix`.

Mec-iS:
thank you @georeth 

I guess what you are looking for is:
```rust
let w: Vec<f64> = Lasso::fit(&x, &y, param)
    .unwrap()
    .coefficients()
    .iterator(0)
    .copied()
    .collect();
```
Please also run `cargo fmt --all -- --check`

georeth:
fixed

PR #344, opened 2025-11-28
add fit_intercept to LASSO

### LINKED ISSUE #342: LASSO Implementation: Condition bug, missing normalization documentation, and intercept fitting option

georeth:
Fixes #342

### Checklist

### Current behaviour
LASSO always fits intercept

### New expected behaviour
add `fit_intercept` option

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,11 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [Unreleased]
+- WARNING: Breaking changes!
+- `LassoParameters` and `LassoSearchParameters` have a new field `fit_intercept`. When it is set to false, the `beta_0` term in the formula will be forced to zero, and `intercept` field in `Lasso` will be set to `None`.
+
+
 ## [0.4.0] - 2023-04-05
 
 ## Added
--- a/src/linear/elastic_net.rs
+++ b/src/linear/elastic_net.rs
@@ -345,6 +345,7 @@ impl<TX: FloatNumber + RealNumber, TY: Number, X: Array2<TX>, Y: Array1<TY>>
                 l1_reg * gamma,
                 parameters.max_iter,
                 TX::from_f64(parameters.tol).unwrap(),
+                true,
             )?;
 
             for i in 0..p {
@@ -371,6 +372,7 @@ impl<TX: FloatNumber + RealNumber, TY: Number, X: Array2<TX>, Y: Array1<TY>>
                 l1_reg * gamma,
                 parameters.max_iter,
                 TX::from_f64(parameters.tol).unwrap(),
+                true,
             )?;
 
             for i in 0..p {
--- a/src/linear/lasso.rs
+++ b/src/linear/lasso.rs
@@ -128,8 +128,8 @@ impl<TX: FloatNumber + RealNumber, TY: Number, X: Array2<TX>, Y: Array1<TY>>
 {
     fn new() -> Self {
         Self {
-            coefficients: Option::None,
-            intercept: Option::None,
+            coefficients: None,
+            intercept: None,
             _phantom_ty: PhantomData,
             _phantom_y: PhantomData,
         }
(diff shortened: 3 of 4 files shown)


georeth:
I've checked it converges to the desired value by comparing with sklearn's LassoLars. (the existing test dataset is strong enough that coordinate descent (sklearn's Lasso) converges slow)

I am not very confident about whether the optimizer code assumes `mean(y) == 0` when it computes search direction.

Mec-iS:
if you are not sure, add a couple of tests. You can just translate the ones in sklearn.

georeth:
There are 3 LASSO algorithms involved:
- coordinate descent: fast if data is well-conditioned, but converge slowly otherwise. (sklearn's Lasso)
- LassoLars: compute exact solution, but has numerical issue in rare corner cases, because it needs to compare floating numbers.
- L1LS: Newton interior point method using approximation for speedup (smartcore's implementation)

sklearn's algorithms are not suitable for my use case, and that's why I want to make changes to smartcore.
I tested this PR against sklearn's LassoLars on random-generated data, and the current status of this PR good enough for me.

georeth:
The L1LS paper has no assumption, and it treat LASSO as its special case, so I think it's correct.

Mec-iS:
the docstring at the top of the module `!//` is for noting down these kind of things.
Write there your assumptions with references to the papers like in https://github.com/smartcorelib/smartcore/blob/2bf5f7a1a5db6809ab64c198a8838f773632a148/src/linear/lasso.rs#L4

georeth:
@Mec-iS
The mod doc is really helpful. I had used it to find the L1LS paper:
https://github.com/smartcorelib/smartcore/blob/2bf5f7a1a5db6809ab64c198a8838f773632a148/src/linear/lasso.rs#L20

The doc also referenced a matlab implementation of L1LS (by the paper writer):
https://github.com/smartcorelib/smartcore/blob/2bf5f7a1a5db6809ab64c198a8838f773632a148/src/linear/lasso.rs#L21

The matlab implementation doesn't normalize y at all: https://github.com/cvxgrp/l1_ls/blob/master/l1_ls.m

Mec-iS:
are these changes breaking the API? the existing API will still be accessible or code that exists should be changed to keep using it?

georeth:
It doesn't breaks the API, and it doesn't introduce a numerical diff.
The only diff I can think of is the (de)serialization of LassoParameter, since I introduced a new option.

georeth:
Oh. If the user constructs a LassoParameters directly without using the builder (`default().with_A(...).with_B(...)`), the code breaks.
```
pub struct LassoParameters {
    pub alpha: f64,
    pub normalize: bool,
    pub tol: f64,
    pub max_iter: usize,
    // a new option here
}
```

This also applies to LassoSearchParameters, and LassoSearchParameters doesn't have a builder interface.

georeth:
Should we set intercept to `None` instead of `Some(0.0)` if `fit_intercept` is false?
```rust
pub struct Lasso<TX: FloatNumber + RealNumber, TY: Number, X: Array2<TX>, Y: Array1<TY>> {
    coefficients: Option<X>,
    intercept: Option<TX>,
    _phantom_ty: PhantomData<TY>,
    _phantom_y: PhantomData<Y>,
}
```
I think coefficients and intercept are Option because SupervisedEstimator trait needs a `new()`.

Mec-iS:
0.0 is a value
None is NaN/null
So pick the right one for this case.
If the value is not used/not relevant, then use None.

Breaking changes need to be added to the CHANGELOG.md file

Mec-iS (approved):

Mec-iS:
Thanks for your contribution

PR #360, opened 2026-03-16
Predict_proba functionality to Random Forest Classifier

skywardfire1:
Checklist
- [yes] My branch is up-to-date with development branch.
- [yes] Everything works and tested on latest stable Rust.
- [yes] Coverage and Linting have been applied 

This PR adds:
1. Proper probability estimation to Decision Tree Classifier, matching scikit-learn's `predict_proba` behavior.
2. Predict_proba functionality to Random Forest Classifier, also in scikit-learn style, which was the goal.

Previously, probability predictions from Decision Tree returned one-hot encoded vectors (1.0 for the majority class, 0.0 for others), which did not reflect actual class distributions in leaf nodes.
While we could still use it in Random Forest, this approach would not provide calibrated probability estimates.

Changes:
1. Node structure extended: Added `class_distribution: Vec<usize>` field to store the class histogram in each node. This data was already being computed during training but was not persisted.
2. DecisionTreeClassifier: Added `predict_proba_for_row_real()` method that returns proper probability distributions based on leaf class counts. The original `predict_proba()` method remains unchanged for backward compatibility. Hope, we will obsolete it one day.
3. RandomForestClassifier: Added public `predict_proba()` method that averages probability distributions from all trees (scikit-learn style), rather than averaging hard class predictions.
4. Testing: Added 5 new tests covering:
   - Probability distributions summing to 1.0
   - Correct class ordering in predictions
   - Mixed-class leaf handling
   - Forest-level probability averaging
5. Debug assertions: Added 3 `debug_assert_eq!`, just to feel better.

Backward Compatibility:
No breaking changes. All existing APIs remain intact.

Note:
Current predict_proba function returns `Vec<Vec<f64>>`, not `DenseMatrix<f64>`, since I didn't find any examples on what is the default behavior or standard for this.

### DIFF
--- a/.github/CONTRIBUTING.md
+++ b/.github/CONTRIBUTING.md
@@ -50,7 +50,7 @@ $ rust-code-analysis-cli -p src/algorithm/neighbour/fastpair.rs --ls 22 --le 213
 
 1. After a PR is opened maintainers are notified
 2. Probably changes will be required to comply with the workflow, these commands are run automatically and all tests shall pass:
-    * **Formatting**: run `rustfmt src/*.rs` to apply automatic formatting
+    * **Formatting**: run `cargo fmt --all -- --check` to apply automatic formatting
     * **Linting**: `clippy` is used with command `cargo clippy --all-features -- -Drust-2018-idioms -Dwarnings`
     * **Coverage** (optional): `tarpaulin` is used with command `cargo tarpaulin --out Lcov --all-features -- --test-threads 1`
     * **Testing**: multiple test pipelines are run for different targets
--- a/src/algorithm/neighbour/cover_tree.rs
+++ b/src/algorithm/neighbour/cover_tree.rs
@@ -290,7 +290,10 @@ impl<T: Debug + PartialEq, D: Distance<T>> CoverTree<T, D> {
             self.new_leaf(p)
         } else {
             let max_dist = self.max(point_set);
-            let next_scale = (max_scale - 1).min(self.get_scale(max_dist));
+            let next_scale = max_scale // bugfix i64::MIN - 1 causes overflow
+                .checked_sub(1) // returns None if overflow
+                .map(|s| s.min(self.get_scale(max_dist)))
+                .unwrap_or(i64::MIN); // safely getting required value
             if next_scale == i64::MIN {
                 let mut children: Vec<Node> = Vec::new();
                 let mut leaf = self.new_leaf(p);
--- a/src/ensemble/base_forest_regressor.rs
+++ b/src/ensemble/base_forest_regressor.rs
@@ -161,25 +161,31 @@ impl<TX: Number + FloatNumber + PartialOrd, TY: Number, X: Array2<TX>, Y: Array1
     /// Predict OOB classes for `x`. `x` is expected to be equal to the dataset used in training.
     pub fn predict_oob(&self, x: &X) -> Result<Y, Failed> {
         let (n, _) = x.shape();
-        if self.samples.is_none() {
-            Err(Failed::because(
-                FailedError::PredictFailed,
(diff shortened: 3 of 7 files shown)


skywardfire1:
applied auto formatting.

Mec-iS:
wow. thank you. will take a look asap

skywardfire1:
some additional info. This is how it looks like in my project

```rust
labels - True:      [1, 1, 4, 0, 0, 5, 3, 0, 0, 5, 0, 0, 0, 0, 0, 4, 4, 1, 2, 3, 5, 1, 4, 0, 3, 1, 0, 0, 3, 3, 0, 0, 0, 4, 5, 1, 1, 0, 0, 1, 5, 2, 4, 4, 0, 0, 1, 1, 3, 4, 0, 0, 4, 2, 2, 3, 4, 5, 5, 0, 0, 5, 0, 0, 0, 4, 4, 1, 5]
labels - Predicted: [5, 1, 4, 0, 0, 5, 2, 0, 0, 5, 0, 0, 4, 0, 0, 4, 4, 5, 2, 3, 5, 5, 4, 0, 3, 3, 0, 0, 3, 3, 0, 0, 0, 4, 5, 1, 1, 0, 0, 1, 5, 2, 4, 1, 0, 0, 1, 5, 3, 4, 0, 0, 4, 4, 4, 3, 4, 5, 5, 0, 0, 5, 0, 0, 0, 4, 4, 1, 1]

Primary probabilities (first 5 samples):
    [[0.0, 0.2, 0.0, 0.0, 0.0, 0.8, 0.0, 0.0],
     [0.0, 0.9, 0.0, 0.0, 0.0, 0.1, 0.0, 0.0],
     [0.1, 0.0, 0.0, 0.0, 0.9, 0.0, 0.0, 0.0],
     [0.9, 0.0, 0.0, 0.0, 0.1, 0.0, 0.0, 0.0],
     [1.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0]] 
```

And [1.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0] result is funny to me since the answer is correct, but __so__ shy

Mec-iS:
this looks OK.

it would be nice to have a test using the iris dataset like this one:
```
    #[cfg_attr(
        all(target_arch = "wasm32", not(target_os = "wasi")),
        wasm_bindgen_test::wasm_bindgen_test
    )]
    #[test]
    fn fit_predict_iris_oob() { 
        ...
```
so the results can be checked on known results.

Mec-iS:
maybe a little better error handling like:
```
  pub fn predict_proba(&self, x: &X) -> Result<Vec<Vec<f64>>, Failed> {                                                 
      let (n, _) = x.shape();                                                                                           
      let mut result = Vec::with_capacity(n);                                                                           
      for i in 0..n {                                                                                                   
          result.push(self.predict_proba_for_row(x, i));                                                                
      }                                                                                                                 
      Ok(result)                                                                                                        
  }  
  ```
  Rest looks good.

Mec-iS (commented):
check out my comments

skywardfire1:
I revisited the tests. As said in commit comment, now there are 2 tests. The first one uses Iris dataset from the beginning of the file, and performs 4 checks.

Everything builds and works perfectly at my side, clippy and fmt --all shows no issues, so I have no idea why builds fail.

What about error handling. I can physically add error checks but it seems useless since there are no operations could possibly return an error if the user doesn't break the API.

skywardfire1:
still, checks don't look good. I offer to throw this PR away, I'll make another one soon after, with same functionality.

Mec-iS:
no problem. proceed as you find right

PR #371, opened 2026-05-20
Replaced unsafe ptr logic with chained split_at_mut in DenseMatrix an…

### LINKED ISSUE #368: tech-debt: replace unsafe raw-pointer iterator_mut in DenseMatrix (basic/matrix.rs) with safe split_at_mut approach
## Context

`DenseMatrix::iterator_mut` in `src/linalg/basic/matrix.rs` uses raw pointer arithmetic with `unsafe` blocks to implement both row-major (axis-0) and column-major (axis-1) mutable iteration:

```rust
// Current implementation (DenseMatrix and DenseMatrixMutView)
let ptr = self.values.as_mut_ptr();
// ...
Box::new((0..nrows).flat_map(move |r| {
    (0..ncols).map(move |c| unsafe {
        &mut *ptr.add(if column_major { r + c * nrows } else { r * ncols + c })
    })
}))
```

This is a **pre-existing pattern** that was not introduced by any recent PR. It is guarded by a `#[cfg(debug_assertions)]` `HashSet`-based aliasing check, and a `// Safety:` comment documents the invariant. However, per the project design principles:

> No use of `unsafe` blocks.

## What Should Be Done

Replace the raw pointer approach with a **safe Rust alternative**. Two viable approaches:

### Option A — `split_at_mut` chain (zero-copy, no allocation)

For row-major (axis-0), the flat `values` Vec is already in the correct order for `DenseMatrix` with `column_major = false`, so `values.iter_mut()` works directly. For column-major storage or axis-1 traversal, use recursive `split_at_mut` to produce non-aliasing `&mut T` slices:

```rust
// Conceptual sketch for column-major axis-0 on a column_major=true DenseMatrix
// Stride = nrows; each column lives at values[col*nrows .. (col+1)*nrows]
let chunks: Vec<&mut T> = self.values
    .chunks_mut(nrows)              // each chunk = one column
    .flat_map(|col| col.iter_mut()) // row-major within each column
    .collect();
Box::new(chunks.into_iter())
```

For axis-1 on a row-major layout the chunks would be of size `ncols`.

### Option B — Restructure to avoid axis-1 mutable iteration

Audit all call sites of `iterator_mut(1)`. If they can be replaced by reading with `iterator(0)` and writing back via `set()`, the mutable column-major iterator can be removed from the trait entirely, simplifying the implementation surface.

## Acceptance Criteria

- [ ] `iterator_mut` for `DenseMatrix` contains no `unsafe` block
- [ ] `iterator_mut` for `DenseMatrixMutView` contains no `unsafe` block  
- [ ] All existing `test_iter_mut` tests continue to pass
- [ ] No new allocations introduced for the hot path (axis-0 row-major case)
- [ ] `cargo clippy --all-features -- -Drust-2018-idioms -Dwarnings` passes clean

## Related

- PR #367 fixed the analogous issue in `src/linalg/ndarray/matrix.rs` using `axis_iter_mut(Axis(1))` (ndarray's safe built-in). The `DenseMatrix` case is slightly harder because it owns a flat `Vec<T>` rather than an ndarray buffer, so the solution requires a different technique.
- The `#[cfg(debug_assertions)]` aliasing check in the current code can be removed once the `unsafe` is gone.

jackpots28:
…d DenseMatrixMutView

Fixes #368 

### Checklist
- [ x ] My branch is up-to-date with development branch.
- [ x ] Everything works and tested on latest stable Rust.
- [ x ] Coverage and Linting have been applied 

### Current behaviour

Changing DenseMatrix and DenseMatrixMutView that uses unsafe ptr arithmetic for generating mutable iterators

### New expected behavior
DenseMatrix and DenseMatrixMutView continue to produce mutable iterators, but uses a "head" / "tail" method + split_at_mut function for iterating
- no test changes required

### Change logs

- Replaced `unsafe` pointer arithmetic in `DenseMatrix` / `DenseMatrixMutView` mutable iterators with a safe, chained `split_at_mut` implementation to ensure memory safety without performance loss.

### DIFF
--- a/src/linalg/basic/matrix.rs
+++ b/src/linalg/basic/matrix.rs
@@ -1073,12 +1073,172 @@ mod tests {
             &[1. + f32::EPSILON, 2., 3.],
             &[4., 5., 6. + f32::EPSILON],
         ])
-        .unwrap();
+            .unwrap();
         let d = DenseMatrix::from_2d_array(&[&[1. + 0.5, 2., 3.], &[4., 5., 6. + f32::EPSILON]])
             .unwrap();
 
         assert!(!relative_eq!(a, b));
         assert!(!relative_eq!(a, d));
         assert!(relative_eq!(a, c));
+
+        let a_int = DenseMatrix::from_2d_array(&[&[1, 2], &[3, 4]]).unwrap();
+        let b_int = DenseMatrix::from_2d_array(&[&[1, 2], &[3, 4]]).unwrap();
+        let c_int = DenseMatrix::from_2d_array(&[&[5, 6], &[7, 8]]).unwrap();
+        assert_eq!(a_int, b_int);
+        assert_ne!(a_int, c_int);
+    }
+
+    #[test]
+    fn test_abs_diff_eq() {
+        let a = DenseMatrix::from_2d_array(&[&[1.0, 2.0], &[3.0, 4.0]]).unwrap();
+        let b = DenseMatrix::from_2d_array(&[&[1.0, 2.0], &[3.0, 4.0000001]]).unwrap();
+        let c = DenseMatrix::from_2d_array(&[&[1.0, 2.0], &[3.0, 4.1]]).unwrap();
+
+        assert!(a.abs_diff_eq(&b, 1e-6));
+        assert!(!a.abs_diff_eq(&c, 1e-6));
+    }
+
+    #[test]
+    fn test_relative_eq() {
+        let a = DenseMatrix::from_2d_array(&[&[1.0, 2.0], &[3.0, 4.0]]).unwrap();
+        let b = DenseMatrix::from_2d_array(&[&[1.0, 2.0], &[3.0, 4.0000001]]).unwrap();
+
+        assert!(relative_eq!(a, b, epsilon = 1e-6, max_relative = 1e-6));
+    }
+
+    #[test]
+    fn test_new_error() {
+        let result = DenseMatrix::new(2, 2, vec![1, 2, 3], true);
+        assert!(result.is_err());
+    }
+
+    #[test]
+    fn test_mut_array_iterator_mut_all_cases() {
+        // Case B: column-major, axis 1 (col-by-col)
+        let mut m1 = DenseMatrix::new(2, 2, vec![1, 2, 3, 4], true).unwrap();
+        m1.iterator_mut(1).for_each(|v| *v += 1);
+        assert_eq!(m1.values, vec![2, 3, 4, 5]);
+
+        // Case A: column-major, axis 0 (row-by-row)
+        let mut m2 = DenseMatrix::new(2, 2, vec![1, 2, 3, 4], true).unwrap();
+        let vals: Vec<i32> = m2.iterator_mut(0).map(|v| *v).collect();
+        assert_eq!(vals, vec![1, 3, 2, 4]);
+        m2.iterator_mut(0).for_each(|v| *v *= 2);
+        assert_eq!(m2.values, vec![2, 4, 6, 8]);
+
+        // Case C: row-major, axis 0 (row-by-row)
+        let mut m3 = DenseMatrix::new(2, 2, vec![1, 2, 3, 4], false).unwrap();
+        m3.iterator_mut(0).for_each(|v| *v += 1);
+        assert_eq!(m3.values, vec![2, 3, 4, 5]);
+
+        // Case D: row-major, axis 1 (col-by-col)
+        let mut m4 = DenseMatrix::new(2, 2, vec![1, 2, 3, 4], false).unwrap();
+        let vals: Vec<i32> = m4.iterator_mut(1).map(|v| *v).collect();
+        assert_eq!(vals, vec![1, 3, 2, 4]);
+        m4.iterator_mut(1).for_each(|v| *v *= 2);
+        assert_eq!(m4.values, vec![2, 4, 6, 8]);
+    }
+
+    #[test]
+    fn test_dense_matrix_mut_view_iter_mut_all_cases() {
+        // Case B: column-major, axis 1 (col-by-col)
+        let mut m1 = DenseMatrix::new(3, 3, (1..10).collect(), true).unwrap();
+        {
+            let mut v = DenseMatrixMutView::new(&mut m1, 0..2, 0..2).unwrap();
+            v.iter_mut(1).for_each(|v| *v = 0);
+        }
+        assert_eq!(m1.values, vec![0, 0, 3, 0, 0, 6, 7, 8, 9]);
+
+        // Case A: column-major, axis 0 (row-by-row)
+        let mut m2 = DenseMatrix::new(3, 3, (1..10).collect(), true).unwrap();
+        {
+            let mut v = DenseMatrixMutView::new(&mut m2, 0..2, 0..2).unwrap();
+            let vals: Vec<i32> = v.iter_mut(0).map(|v| *v).collect();
+            assert_eq!(vals, vec![1, 4, 2, 5]);
+            v.iter_mut(0).for_each(|v| *v = 0);
+        }
+        assert_eq!(m2.values, vec![0, 0, 3, 0, 0, 6, 7, 8, 9]);
+
+        // Case C: row-major, axis 0 (row-by-row)
+        let mut m3 = DenseMatrix::new(3, 3, (1..10).collect(), false).unwrap();
+        {
+            let mut v = DenseMatrixMutView::new(&mut m3, 0..2, 0..2).unwrap();
+            v.iter_mut(0).for_each(|v| *v = 0);
+        }
+        assert_eq!(m3.values, vec![0, 0, 3, 0, 0, 6, 7, 8, 9]);
+
+        // Case D: row-major, axis 1 (col-by-col)
+        let mut m4 = DenseMatrix::new(3, 3, (1..10).collect(), false).unwrap();
+        {
+            let mut v = DenseMatrixMutView::new(&mut m4, 0..2, 0..2).unwrap();
+            let vals: Vec<i32> = v.iter_mut(1).map(|v| *v).collect();
+            assert_eq!(vals, vec![1, 4, 2, 5]);
+            v.iter_mut(1).for_each(|v| *v = 0);
+        }
+        assert_eq!(m4.values, vec![0, 0, 3, 0, 0, 6, 7, 8, 9]);
+    }
+
+    #[test]
+    fn test_is_empty() {
+        let m = DenseMatrix::new(2, 2, vec![1, 2, 3, 4], true).unwrap();
+        assert!(!m.is_empty());
+        let empty: DenseMatrix<i32> = DenseMatrix::new(0, 0, vec![], true).unwrap();
+        assert!(empty.is_empty());
+    }
+
+    #[test]
+    fn test_stride_range_error() {
+        let _m = DenseMatrix::new(2, 2, vec![1, 2, 3, 4], true).unwrap();
+    }
+
+    #[test]
+    #[should_panic(expected = "Invalid index (2,0) for 2x2 matrix")]
+    fn test_get_out_of_bounds() {
+        let m = DenseMatrix::from_2d_array(&[&[1, 2], &[3, 4]]).unwrap();
+        m.get((2, 0));
+    }
+
+    #[test]
+    fn test_transpose_row_major() {
+        let m = DenseMatrix::new(2, 3, vec![1, 2, 3, 4, 5, 6], false).unwrap();
+        let mt = m.transpose();
+        assert!(mt.column_major);
+        assert_eq!(mt.nrows, 3);
+        assert_eq!(mt.ncols, 2);
+        assert_eq!(mt.values, vec![1, 2, 3, 4, 5, 6]);
+    }
+
+    #[test]
+    #[should_panic(expected = "For two dimensional array `axis` should be either 0 or 1")]
+    fn test_iterator_invalid_axis() {
+        let m = DenseMatrix::from_2d_array(&[&[1, 2], &[3, 4]]).unwrap();
+        let _ = m.iterator(2);
+    }
+
+    #[test]
+    #[should_panic(expected = "For two dimensional array `axis` should be either 0 or 1")]
+    fn test_iterator_mut_invalid_axis() {
+        let mut m = DenseMatrix::from_2d_array(&[&[1, 2], &[3, 4]]).unwrap();
+        let _ = m.iterator_mut(2);
+    }
+
+    #[test]
+    fn test_view_1d_access() {
+        let m = DenseMatrix::from_2d_array(&[&[1, 2, 3], &[4, 5, 6]]).unwrap();
+        let v_row = DenseMatrixView::new(&m, 0..1, 0..3).unwrap();
+        assert_eq!(<DenseMatrixView<'_, i32> as Array<i32, usize>>::shape(&v_row), 3);
+        assert_eq!(<DenseMatrixView<'_, i32> as Array<i32, usize>>::get(&v_row, 1), &2);
+
+        let v_col = DenseMatrixView::new(&m, 0..2, 1..2).unwrap();
+        assert_eq!(<DenseMatrixView<'_, i32> as Array<i32, usize>>::shape(&v_col), 2);
+        assert_eq!(<DenseMatrixView<'_, i32> as Array<i32, usize>>::get(&v_col, 1), &5);
+    }
+
+    #[test]
+    #[should_panic(expected = "This is neither a column nor a row")]
+    fn test_view_1d_access_invalid() {
+        let m = DenseMatrix::from_2d_array(&[&[1, 2, 3], &[4, 5, 6]]).unwrap();
+        let v = DenseMatrixView::new(&m, 0..2, 0..2).unwrap();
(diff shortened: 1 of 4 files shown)


Mec-iS:
thank you!

please run `cargo fmt --all -- --check` and the other checks needed.

jackpots28:
I updated the formatting and caught the issue with sort_by that clippy complained about. Test coverage is still slightly lower "src/linalg/basic/matrix.rs: 225/364" | "59.72% coverage, 5238/8771 lines covered", should I add some tests for more coverage?

Mec-iS:
> I updated the formatting and caught the issue with sort_by that clippy complained about. Test coverage is still slightly lower "src/linalg/basic/matrix.rs: 225/364" | "59.72% coverage, 5238/8771 lines covered", should I add some tests for more coverage?

adding more tests on the methods affected by changes is always welcome 🐳 thanks. at your convenience.

jackpots28:
> > I updated the formatting and caught the issue with sort_by that clippy complained about. Test coverage is still slightly lower "src/linalg/basic/matrix.rs: 225/364" | "59.72% coverage, 5238/8771 lines covered", should I add some tests for more coverage?
> 
> adding more tests on the methods affected by changes is always welcome 🐳 thanks. at your convenience.

Rolled through and updated matrix.rs to include 33 new cases

src/linalg/basic/matrix.rs
```diff
@@ -143,81 +143,135 @@ impl<'a, T: Debug + Display + Copy + Sized> DenseMatrixMutView<'a, T> {
...
-                            );
-                        }
+        // Axis = 0: row-by-row (outer loop over rows, inner over cols)
+        // Axis = 1: col-by-col (outer loop over cols, inner over rows)
+        // Four cases: column-major (axis 0 or 1), row-major (axis 1 or 0)
+
+        // Collect all mutable references up-front using split_at_mut so
+        // that the resulting iterator owns no borrow of "self.values"
+
+        match (column_major, axis) {
```
genefold-ai: 🔴 **[HIGH] Potential panic: `col_slice[..nrows]` when `stride < nrows`**

In `DenseMatrixMutView`, when the view has a stride smaller than `nrows` (e.g. a sub-view into a larger matrix where padding or offset applies), `split_at_mut(stride)` will produce a `col_slice` shorter than `nrows`, causing `col_slice[..nrows]` to panic at runtime.

The original `unsafe` implementation handled this correctly via direct offset arithmetic; this safe version does not inherit that safety.

Please add an explicit guard before slicing:

```rust
assert!(
    col_slice.len() >= nrows,
    "iter_mut: stride ({stride}) < nrows ({nrows}): view layout is inconsistent"
);
for elem in col_slice[..nrows].iter_mut() {
```

Or alternatively, use `.get_mut(..nrows)` and propagate/handle the `None` case rather than panicking silently later.

src/linalg/basic/matrix.rs
```diff
@@ -143,81 +143,135 @@ impl<'a, T: Debug + Display + Copy + Sized> DenseMatrixMutView<'a, T> {
...
+        assert!(
+            axis == 1 || axis == 0,
+            "For two dimensional array `axis` should be either 0 or 1"
+        );
+
         let column_major = self.column_major;
         let stride = self.stride;
         let nrows = self.nrows;
         let ncols = self.ncols;
-        let ptr = self.values.as_mut_ptr();
```
genefold-ai, line 175: 🔴 **[HIGH] Subtraction underflow: `ncols - 1` panics (debug) or wraps (release) when `ncols == 0`**

The expression `if _c == ncols - 1` is evaluated without first checking whether `ncols` is zero. When `ncols == 0`:
- In **debug mode**: panics with a subtraction overflow.
- In **release mode**: wraps around to `usize::MAX`, causing the branch condition to never match and likely producing an incorrect or out-of-bounds slice.

This can be triggered by a zero-column `DenseMatrixMutView` (e.g. a view into an empty region). The same issue exists symmetrically for `nrows - 1` in the row-major cases.

Please add an early guard at the top of the method:

```rust
if ncols == 0 || nrows == 0 {
    return Box::new(std::iter::empty());
}
```

This also simplifies all subsequent arithmetic by making zero-size cases unreachable.

src/linalg/basic/matrix.rs
```diff
@@ -143,81 +143,135 @@ impl<'a, T: Debug + Display + Copy + Sized> DenseMatrixMutView<'a, T> {
...
-            0 => Box::new((0..nrows).flat_map(move |r| {
-                (0..ncols).map(move |c| unsafe {
-                    &mut *ptr.add(if column_major {
-                        r + c * stride
-                    } else {
-                        r * stride + c
+                let mut indexed: Vec<(usize, &'b mut T)> = by_col
+                    .into_iter()
+                    .enumerate()
+                    .map(|(flat_col_idx, r)| {
```
genefold-ai: 🟡 **[MEDIUM] Triple `Vec` allocation in Case A is redundant — `refs` can be eliminated**

In Case A (column-major, row-by-row) the code allocates `by_col`, then `indexed`, then a final `refs` Vec that is immediately consumed by `into_iter()`:

```rust
let mut refs: Vec<&'b mut T> = Vec::with_capacity(total);
refs.extend(indexed.into_iter().map(|(_, r)| r));
Box::new(refs.into_iter())
```

The final `refs` is a pure copy of the `indexed` values with keys stripped. It can be removed entirely:

```rust
Box::new(indexed.into_iter().map(|(_, r)| r))
```

This saves one `Vec` allocation and one full iteration pass per `iter_mut` call. The same pattern appears in Case D of this method and in both Cases A and D of `DenseMatrix::iterator_mut` — four sites in total should be cleaned up.

src/linalg/basic/matrix.rs
```diff
@@ -143,81 +143,135 @@ impl<'a, T: Debug + Display + Copy + Sized> DenseMatrixMutView<'a, T> {
     }
```
genefold-ai, line 158: 🟡 **[MEDIUM] Duplicated `split_at_mut` chunking logic across cases — extract a private helper**

Cases B and C in `DenseMatrixMutView::iter_mut`, and their analogues in `DenseMatrix::iterator_mut`, share nearly identical code for iterating over a strided slice with `split_at_mut`:

```rust
let mut remaining: &'b mut [T] = self.values;
for _c in 0..ncols {
    let col_end = if _c == ncols - 1 { remaining.len() } else { stride };
    let (col_slice, tail) = remaining.split_at_mut(col_end);
    for elem in col_slice[..nrows].iter_mut() { refs.push(elem); }
    remaining = tail;
}
```

This 9-line block is repeated (with minor parameter changes) four times across the two impls. Consider extracting a private helper:

```rust
/// Collects mutable references to the first `take` elements of each
/// chunk of size `chunk` from `slice`, advancing through the tail.
fn collect_strided_mut<'a, T>(
    slice: &'a mut [T],
    chunks: usize,
    chunk_size: usize,
    take: usize,
) -> Vec<&'a mut T> {
    let mut refs = Vec::with_capacity(chunks * take);
    let mut remaining = slice;
    for i in 0..chunks {
        let end = if i == chunks - 1 { remaining.len() } else { chunk_size };
        let (head, tail) = remaining.split_at_mut(end);
        refs.extend(head[..take].iter_mut());
        remaining = tail;
    }
    refs
}
```

This reduces duplication, makes the zero-size guard easier to add in one place, and makes the bounds check (`head.len() >= take`) easier to enforce uniformly.

src/linalg/basic/matrix.rs
```diff
@@ -502,49 +556,84 @@ impl<T: Debug + Display + Copy + Sized> MutArray<T, (usize, usize)> for DenseMat
...
+                    .values
+                    .chunks_mut(ncols)
+                    .flat_map(|row| row.iter_mut())
+                    .collect();
+                Box::new(refs.into_iter())
+            }
+
+            // Case D: row-major, col-by-col
+            (false, _) => {
+                let total = nrows * ncols;
```
genefold-ai: 🟡 **[MEDIUM] O(n log n) sort for cross-axis traversal — consider direct index computation**

Cases A and D both use this pattern to reorder elements for cross-axis iteration:

```rust
let mut indexed: Vec<(usize, &'b mut T)> = by_col
    .into_iter()
    .enumerate()
    .map(|(flat_col_idx, elem)| {
        let c = flat_col_idx / nrows;
        let r = flat_col_idx % nrows;
        (r * ncols + c, elem)
    })
    .collect();
indexed.sort_unstable_by_key(|(idx, _)| *idx);
```

This performs an O(n log n) sort on `n = nrows * ncols` elements just to reorder them into row-major (or col-major) access order. For large matrices this is a meaningful overhead compared to the original O(n) pointer arithmetic.

For `DenseMatrix` (contiguous, non-strided allocation), the same reordering can be done directly using `chunks_mut` with a transposition index formula, which is O(n):

```rust
// Case A: column-major, iterate row-by-row (no sort needed)
// Conceptually: output[r * ncols + c] = col_major_data[c * nrows + r]
// This is just a transpose read-order, achievable with two nested iterators.
```

If a sort is kept for correctness in the strided `DenseMatrixMutView` case, please add a comment explaining why direct index computation is not viable there.

genefold-ai (changes requested):
The security goal is right and the overall direction is solid — removing `unsafe` pointer arithmetic is a clear win for long-term maintainability. Two runtime bugs need to be fixed before merging (see inline comments marked 🔴), and there are three medium-priority cleanup items (🟡) that would meaningfully improve performance and reduce duplication. Happy to discuss any of these — great contribution.

Mec-iS:
@jackpots28 please see the automated code-review by @genefold-ai and feel free to decide which patches to apply

Mec-iS:
This one more highlighted by another automated code-review (may overlap the previous ones):

🔴 [HIGH] Bug 1: Subtraction underflow when ncols == 0 or nrows == 0

src/linalg/basic/matrix.rs — both DenseMatrixMutView::iter_mut and DenseMatrix::iterator_mut
The expression _c == ncols - 1 (and symmetrically _r == nrows - 1 in the row-major cases) is evaluated without 
checking whether ncols or nrows is zero:
```
for _c in 0..ncols {
    let col_end = if _c == ncols - 1 {  // panics when ncols == 0
        remaining.len()
    } else {
        stride
    };
```

When ncols == 0:
  • Debug builds: Panics with attempt to subtract with overflow.
  • Release builds: Wraps to usize::MAX, the condition never matches, and split_at_mut(usize::MAX) panics or produces 
garbage.

This was correctly pointed out by the genefold-ai review. Add an early guard:
```
if ncols == 0 || nrows == 0 {
    return Box::new(std::iter::empty());
}
```

About the tests:
| 🟢 Low | Missing edge case tests | Add tests for empty/1×N/N×1 cases |

Also I am trying https://docs.rs/cargo-crap/latest/cargo_crap/ to analyse complexity/risk patterns.

Mec-iS:
Not too bad
```
$ cargo crap --workspace --lcov lcov.info --summary
warning: 13 source files had no matching entry in the LCOV report — verify your --lcov path or coverage tool configuration:
/smartcore/src/dataset/boston.rs
/smartcore/src/dataset/breast_cancer.rs
/smartcore/src/dataset/diabetes.rs
/smartcore/src/dataset/digits.rs
/smartcore/src/dataset/generator.rs
/smartcore/src/dataset/iris.rs
/smartcore/src/dataset/mod.rs
/smartcore/src/linalg/ndarray/matrix.rs
/smartcore/src/linalg/ndarray/vector.rs
/smartcore/src/model_selection/hyper_tuning/grid_search.rs
/smartcore/src/readers/csv.rs
/smartcore/src/readers/error.rs
/smartcore/src/readers/io_testing.rs
Per-crate summary:
┌───────────┬───────────┬────────┐
│ Crate     ┆ Functions ┆ Crappy │
╞═══════════╪═══════════╪════════╡
│ smartcore ┆       917 ┆     16 │
└───────────┴───────────┴────────┘
✗ Analyzed: 917 · Crappy: 16 (threshold 30) · Worst: CategoricalNBDistribution::eq (CRAP 182.0)
```

Mec-iS:
this is a problem, can't merge with this as it breaks lazy evaluation:  

## 1. Significant Performance Regression — Eager Collection

src/linalg/basic/matrix.rs
Lines: Both iter_mut implementations
The original code returned lazy iterators using flat_map and pointer arithmetic. The new implementation eagerly 
collects all mutable references into a Vec before returning an iterator:
```
// New implementation pattern
for i in 0..cols {
    // ... split_at_mut logic ...
    refs.extend(col_slice.iter_mut());  // Collects all refs upfront
}
Box::new(refs.into_iter())  // Returns iterator over collected Vec
```

Problems:
  • Memory overhead: Allocates a full Vec of ncols mutable references even for simple iteration
  • Latency: Must traverse entire matrix before yielding first element
  • Breaks iterator composability: Callers expecting lazy evaluation (e.g., .take(5).collect()) now pay full cost

Recommendation: Use flat_map with split_at_mut to maintain lazy evaluation:
```
fn iter_mut<'b>(&'b mut self, axis: u8) -> Box<dyn Iterator<Item = &'b mut T> + 'b> {
    if axis == 0 {
        Box::new(
            self.values
                .chunks_mut(self.stride)
                .take(ncols)
                .flat_map(move |col| col[..nrows].iter_mut())
        )
    } else {
        // Similar for axis == 1
    }
}
```

also code duplication:
## 3. Code Duplication Across 4 Cases

src/linalg/basic/matrix.rs
Lines: DenseMatrixMutView::iter_mut and DenseMatrix::iterator_mut
Both methods have nearly identical logic for handling:
  • Column-major vs row-major storage
  • Axis 0 (row-wise) vs axis 1 (column-wise) iteration
  • split_at_mut chunking with special case for last column

This results in ~200 lines of duplicated logic.
Recommendation: Extract a shared helper function:
```
/// Build mutable iterator over matrix elements in specified axis
fn build_mut_iter<'b, T>(
    values: &'b mut [T],
    nrows: usize,
    ncols: usize,
    stride: usize,
    column_major: bool,
    axis: u8,
) -> Box<dyn Iterator<Item = &'b mut T> + 'b> {
    // Shared implementation
}
```

jackpots28:
> this is a problem, can't merge with this as it breaks lazy evaluation:  
> 
> 
> 
> ## 1. Significant Performance Regression — Eager Collection
> 
> 
> 
> File: src/linalg/basic/matrix.rs
> 
> Lines: Both iter_mut implementations
> 
> The original code returned lazy iterators using flat_map and pointer arithmetic. The new implementation eagerly 
> 
> collects all mutable references into a Vec before returning an iterator:
> 
> ```
> 
> // New implementation pattern
> 
> for i in 0..cols {
> 
>     // ... split_at_mut logic ...
> 
>     refs.extend(col_slice.iter_mut());  // Collects all refs upfront
> 
> }
> 
> Box::new(refs.into_iter())  // Returns iterator over collected Vec
> 
> ```
> 
> 
> 
> Problems:
> 
>   • Memory overhead: Allocates a full Vec of ncols mutable references even for simple iteration
> 
>   • Latency: Must traverse entire matrix before yielding first element
> 
>   • Breaks iterator composability: Callers expecting lazy evaluation (e.g., .take(5).collect()) now pay full cost
> 
> 
> 
> Recommendation: Use flat_map with split_at_mut to maintain lazy evaluation:
> 
> ```
> 
> fn iter_mut<'b>(&'b mut self, axis: u8) -> Box<dyn Iterator<Item = &'b mut T> + 'b> {
> 
>     if axis == 0 {
> 
>         Box::new(
> 
>             self.values
> 
>                 .chunks_mut(self.stride)
> 
>                 .take(ncols)
> 
>                 .flat_map(move |col| col[..nrows].iter_mut())
> 
>         )
> 
>     } else {
> 
>         // Similar for axis == 1
> 
>     }
> 
> }
> 
> ```
> 
> 
> 
> also code duplication:
> 
> ## 3. Code Duplication Across 4 Cases
> 
> 
> 
> File: src/linalg/basic/matrix.rs
> 
> Lines: DenseMatrixMutView::iter_mut and DenseMatrix::iterator_mut
> 
> Both methods have nearly identical logic for handling:
> 
>   • Column-major vs row-major storage
> 
>   • Axis 0 (row-wise) vs axis 1 (column-wise) iteration
> 
>   • split_at_mut chunking with special case for last column
> 
> 
> 
> This results in ~200 lines of duplicated logic.
> 
> Recommendation: Extract a shared helper function:
> 
> ```
> 
> /// Build mutable iterator over matrix elements in specified axis
> 
> fn build_mut_iter<'b, T>(
> 
>     values: &'b mut [T],
> 
>     nrows: usize,
> 
>     ncols: usize,
> 
>     stride: usize,
> 
>     column_major: bool,
> 
>     axis: u8,
> 
> ) -> Box<dyn Iterator<Item = &'b mut T> + 'b> {
> 
>     // Shared implementation
> 
> }
> 
> ```

Yeah, I'm working on the deduplication via some single collector logic to use across the four instances instead. Also got curious and started into some bigger matrices testing, found what you stated with degradation since it does the full collect. I'll work through it!

jackpots28:
I made some updates over the weekend to reduce code duplication and improve performance. Would using Numpy as an equivalent benchmark suffice when doing matrix arithmetic?

Mec-iS:
> I made some updates over the weekend to reduce code duplication and improve performance. Would using Numpy as an equivalent benchmark suffice when doing matrix arithmetic?

yes, results should be inline with numpy/scikit-learn

Mec-iS:
I am closing this considering the relevant updates brought on by v0.6
Please let me know if you want to reopen.

PR #379, opened 2026-07-09
Add getter for parameters in all model structs.

### LINKED ISSUE #378: Add getter methods for model hyperparameters after deserialization
I have noticed that when a model is deserialized, access to the hyperparameters that were used to create the model is no longer available.

This makes it difficult to inspect, debug, log, or reproduce a trained model after loading it from serialized storage. In many workflows, the trained model is saved and later loaded in a separate process, notebook, service, or deployment environment. At that point, it would be useful to retrieve the hyperparameters that were originally used to construct or train the model.

## Current Behavior

After deserializing a model, the model can be used for prediction, but there does not appear to be a public getter or accessor for the hyperparameters used during model creation or training.

For example, if a model was created with parameters such as max_depth, n_trees, criterion, learning_rate, alpha, or similar algorithm-specific settings, those values are not easily accessible from the deserialized model instance.

## Expected Behavior

Models should expose a getter function or equivalent public accessor that returns the hyperparameters used to create/train the model.

For example:

let params = model.parameters();

or:

let params = model.get_parameters();

The returned value could be the same parameter struct used when creating the model, or an equivalent read-only representation.

## Why This Would Be Useful

Adding access to model hyperparameters after deserialization would help with:

Model reproducibility
Debugging production models
Experiment tracking
Auditability
Comparing serialized models
Logging model configuration after loading
Validating that a loaded model matches expected training settings
Suggested Change

Store the model’s hyperparameters as part of the trained model struct and expose them through a public getter method.

Ideally, this should work consistently across supported models where hyperparameters are passed during training.

## Additional Context
Thought to generate this ticket with gpt, my use case falls more under "Experiment Tracking" as it would be helpful to reference the latest hyperparameters used and maybe associated error metrics. I would believe error metrics can be reproduced if a seed is always provided in any case.

HubertMajewski:
Fixes https://github.com/smartcorelib/smartcore/issues/378

### Current behaviour

Several fitted models do not expose the hyperparameters that were used to create or train them.

After a model is fitted, and especially after it is serialized and deserialized, users can inspect learned model state such as coefficients, intercepts, trees, classes, or other fitted attributes where available, but the original training parameters are not consistently accessible from the model instance.

This makes it difficult to:

* inspect a loaded model’s configuration,
* debug model behavior after deserialization,
* log model metadata,
* reproduce training settings,
* compare two fitted models,
* and verify that a loaded model was trained with the expected hyperparameters.

### The follow-up changes in this revision:

* fix the Clippy failure caused by cloning a `Copy` parameter value;
* document the serialization limitation of borrowed `SVC` parameters;
* add `MultiClassSVC::parameters()` for API consistency;
* explain why the affected generic parameter types use manual Serde implementations; and
* remove synthetic tests that only forced private `parameters` fields to `None`, while retaining positive fit/getter coverage and a real SVC serialization test.
* Documents that `SVC::parameters()` returns `None` after deserialization because the fitted model stores a borrowed parameter reference.
* Retains the previously introduced public `SVCParameters::seed` field and public `DecisionTreeClassifier::parameters()` accessor

### New expected behaviour

Fitted models expose a public `.parameters()` getter that returns the hyperparameters used to create or train the model.

The getter provides read-only access to the model’s parameter struct, allowing users to inspect model configuration after fitting and after deserialization where the parameters are stored with the model.

This improves model reproducibility, auditability, debugging, and experiment tracking without requiring users to separately persist parameter structs outside the model.

#### Added
* Added public .parameters() getter methods to fitted model types that store their training hyperparameters.
* Added or updated parameter storage on fitted models where needed so the original training parameters can be inspected after fitting.
* Added support for retrieving model hyperparameters after deserialization where applicable.
* Added parameter round-trip tests that verify the values returned by .parameters() match the values supplied during fitting.

#### Changed
* Exposed DecisionTreeClassifier::parameters() as a public method.
* Exposed SVCParameters::seed as a public field.
* Created custom Deserialize implementation to support deserialization without requiring Default to the public API.

### `SVC` serialization

`SVC` stores its fitted parameters as a borrowed reference:

```rust
parameters: Option<&'a SVCParameters<TX, TY, X, Y>>,
```

Serde cannot reconstruct that borrow, so the field remains skipped during serialization. The `SVC::parameters()` documentation now states explicitly that a deserialized `SVC` returns `None`, and the Serde round-trip test verifies that behavior. Added an assert to the test to expect a `None`.

This revision intentionally does **not** clone `SVCParameters` into the fitted model. Keeping the existing borrowed design avoids adding clone bounds, and potentially non-trivial kernel cloning to the SVM implementation solely for parameter introspection. 

### `MultiClassSVC`

`MultiClassSVC::parameters()` now exposes the parameter value shared by its one-vs-one binary classifiers. All classifiers are fitted from the same borrowed `SVCParameters`, so the getter returns the reference stored by the first fitted classifier. It returns `None` for an unfitted or empty model.

### Manual Serde implementations

The manual `Deserialize` implementations include a maintenance comment explaining that they avoid the implicit `Default` bounds that `#[derive(Deserialize)]` would add to generic parameter types.

The implementations are in:
* `DBSCANParameters<T, D>`
* `LogisticRegressionParameters<T>`
* `BernoulliNBParameters<T>`
* `KNNClassifierParameters<T, D>`
* `KNNRegressorParameters<T, D>`

#### Cluster

* [x] `AgglomerativeClustering`
* [x] `DBSCAN`
* [x] `KMeans`

#### Decomposition

* [x] `PCA`
* [x] `SVD`

#### Ensemble

* [x] `ExtraTreesRegressor`
* [x] `RandomForestClassifier`
* [x] `RandomForestRegressor`

#### Linear

* [x] `ElasticNet`
* [x] `Lasso`
* [x] `LinearRegression`
* [x] `LogisticRegression`
* [x] `RidgeRegression`

#### Naive Bayes

* [x] `BernoulliNB`
* [x] `CategoricalNB`
* [x] `GaussianNB`
* [x] `MultinomialNB`

#### Neighbors

* [x] `KNNClassifier`
* [x] `KNNRegressor`

#### SVM

* [x] `SVC` (Borrowed)
* [x] `MultiClassSVC` (Borrowed)
* [x] `SVR` (Borrowed)

#### Tree

* [x] `DecisionTreeClassifier`
* [x] `DecisionTreeRegressor`

#### XGBoost

* [x] `XGRegressor`

### DIFF
--- a/src/cluster/agglomerative.rs
+++ b/src/cluster/agglomerative.rs
@@ -188,10 +188,9 @@ impl<TX: Number, TY: Number, X: Array2<TX>, Y: Array1<TY>> AgglomerativeClusteri
     /// Getter for parameters used in the model
     ///
     /// # Returns
-    /// Parameters used to setup the model
-    pub fn parameters(&self) -> &AgglomerativeClusteringParameters {
-        assert!(self.parameters.is_some());
-        &self.parameters.as_ref().unwrap()
+    /// `Some` with the parameters used to configure the model, or `None` if unavailable.
+    pub fn parameters(&self) -> Option<&AgglomerativeClusteringParameters> {
+        self.parameters.as_ref()
     }
 }
 
@@ -324,5 +323,36 @@ mod tests {
         );
 
         assert!(result.is_err());
+    }
+
+    #[test]
+    fn test_can_get_assigned_parameters() {
+        let data = vec![0.0, 0.0, 5.0, 5.0, 10.0, 10.0];
+        let matrix = DenseMatrix::new(3, 2, data, false).unwrap();
+        let parameters = AgglomerativeClusteringParameters::default().with_n_clusters(1);
+        let expected_parameters = parameters.clone();
+        let clustering = AgglomerativeClustering::<f64, f64, DenseMatrix<f64>, Vec<f64>>::fit(
+            &matrix, parameters,
+        )
+        .unwrap();
+
+        let actual_parameters = clustering
+            .parameters()
+            .expect("parameters should be set after fitting");
+        assert_eq!(actual_parameters.n_clusters, expected_parameters.n_clusters);
+    }
+
+    #[test]
+    fn test_returns_none_on_no_parameters() {
+        let clustering = AgglomerativeClustering::<f64, f64, DenseMatrix<f64>, Vec<f64>> {
+            labels: Vec::new(),
+            parameters: None,
+            _phantom_tx: PhantomData,
+            _phantom_ty: PhantomData,
+            _phantom_x: PhantomData,
+            _phantom_y: PhantomData,
+        };
+
+        assert!(clustering.parameters().is_none());
     }
 }
--- a/src/cluster/agglomerative.rs
+++ b/src/cluster/agglomerative.rs
@@ -76,6 +76,7 @@ impl Default for AgglomerativeClusteringParameters {
 pub struct AgglomerativeClustering<TX: Number, TY: Number, X: Array2<TX>, Y: Array1<TY>> {
     /// The cluster label assigned to each sample.
     pub labels: Vec<usize>,
+    parameters: Option<AgglomerativeClusteringParameters>,
     _phantom_tx: PhantomData<TX>,
     _phantom_ty: PhantomData<TY>,
     _phantom_x: PhantomData<X>,
@@ -176,11 +177,21 @@ impl<TX: Number, TY: Number, X: Array2<TX>, Y: Array1<TY>> AgglomerativeClusteri
         }
         Ok(AgglomerativeClustering {
             labels,
+            parameters: Some(parameters),
             _phantom_tx: PhantomData,
             _phantom_ty: PhantomData,
             _phantom_x: PhantomData,
             _phantom_y: PhantomData,
         })
+    }
+
+    /// Getter for parameters used in the model
+    ///
+    /// # Returns
+    /// Parameters used to setup the model
+    pub fn parameters(&self) -> &AgglomerativeClusteringParameters {
+        assert!(self.parameters.is_some());
+        &self.parameters.as_ref().unwrap()
     }
 }
 
--- a/src/cluster/dbscan.rs
+++ b/src/cluster/dbscan.rs
@@ -431,7 +431,7 @@ impl<TX: Number, TY: Number, X: Array2<TX>, Y: Array1<TY>, D: Distance<Vec<TX>>>
 
         Ok(result)
     }
-    
+
     /// Getter for parameters used in the model
     ///
     /// # Returns
@@ -557,7 +557,10 @@ mod tests {
             format!("{:?}", actual_parameters.distance),
             format!("{:?}", expected_parameters.distance)
         );
-        assert_eq!(actual_parameters.min_samples, expected_parameters.min_samples);
+        assert_eq!(
+            actual_parameters.min_samples,
+            expected_parameters.min_samples
+        );
         assert_eq!(actual_parameters.eps, expected_parameters.eps);
         assert_eq!(
             std::mem::discriminant(&actual_parameters.algorithm),
@@ -580,20 +583,19 @@ mod tests {
 
         println!("{labels:?}");
     }
-    
+
     #[test]
     fn test_can_get_assigned_parameters() {
         let data = vec![0.0, 0.0, 0.5, 0.5, 10.0, 10.0];
         let matrix = DenseMatrix::new(3, 2, data, false).unwrap();
-        let parameters: DBSCANParameters<f64, Euclidian<f64>> =
-            DBSCANParameters::default().with_eps(1.0).with_min_samples(1);
+        let parameters: DBSCANParameters<f64, Euclidian<f64>> = DBSCANParameters::default()
+            .with_eps(1.0)
+            .with_min_samples(1);
         let expected_parameters = parameters.clone();
-        let clustering =
-            DBSCAN::<f64, f64, DenseMatrix<f64>, Vec<f64>, Euclidian<f64>>::fit(
-                &matrix,
-                parameters,
-            )
-            .unwrap();
+        let clustering = DBSCAN::<f64, f64, DenseMatrix<f64>, Vec<f64>, Euclidian<f64>>::fit(
+            &matrix, parameters,
+        )
+        .unwrap();
 
         let actual_parameters = clustering
             .parameters()
@@ -602,7 +604,10 @@ mod tests {
             format!("{:?}", actual_parameters.distance),
             format!("{:?}", expected_parameters.distance)
         );
-        assert_eq!(actual_parameters.min_samples, expected_parameters.min_samples);
+        assert_eq!(
+            actual_parameters.min_samples,
+            expected_parameters.min_samples
+        );
         assert_eq!(actual_parameters.eps, expected_parameters.eps);
         assert_eq!(
             std::mem::discriminant(&actual_parameters.algorithm),
@@ -614,14 +619,13 @@ mod tests {
     fn test_returns_none_on_no_parameters() {
         let data = vec![0.0, 0.0, 0.5, 0.5, 10.0, 10.0];
         let matrix = DenseMatrix::new(3, 2, data, false).unwrap();
-        let parameters: DBSCANParameters<f64, Euclidian<f64>> =
-            DBSCANParameters::default().with_eps(1.0).with_min_samples(1);
-        let mut clustering =
-            DBSCAN::<f64, f64, DenseMatrix<f64>, Vec<f64>, Euclidian<f64>>::fit(
-                &matrix,
-                parameters,
-            )
-            .unwrap();
+        let parameters: DBSCANParameters<f64, Euclidian<f64>> = DBSCANParameters::default()
+            .with_eps(1.0)
+            .with_min_samples(1);
+        let mut clustering = DBSCAN::<f64, f64, DenseMatrix<f64>, Vec<f64>, Euclidian<f64>>::fit(
+            &matrix, parameters,
+        )
+        .unwrap();
         clustering.parameters = None;
 
         assert!(clustering.parameters().is_none());
--- a/src/cluster/dbscan.rs
+++ b/src/cluster/dbscan.rs
@@ -64,6 +64,7 @@ pub struct DBSCAN<TX: Number, TY: Number, X: Array2<TX>, Y: Array1<TY>, D: Dista
     num_classes: usize,
     knn_algorithm: KNNAlgorithm<TX, D>,
     eps: f64,
+    parameters: Option<DBSCANParameters<TX, D>>,
     _phantom_ty: PhantomData<TY>,
     _phantom_x: PhantomData<X>,
     _phantom_y: PhantomData<Y>,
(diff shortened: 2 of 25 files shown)


Mec-iS (commented):
## Review: Add getter for parameters in all model structs

Thank you for this PR — the motivation is solid and the coverage across 25 files is impressive. However, there are several recurring implementation issues that should be addressed before merging.

---

### 🔴 Critical

#### `assert!` + `unwrap()` is an anti-pattern

Every getter in this PR follows this pattern:

```rust
pub fn parameters(&self) -> &KMeansParameters {
    assert!(self.parameters.is_some());
    &self.parameters.as_ref().unwrap()
}
```

This is redundant: `assert!` panics in debug builds, then `unwrap()` would panic independently anyway. The idiomatic Rust approach is:

```rust
pub fn parameters(&self) -> &KMeansParameters {
    self.parameters.as_ref().expect("parameters not set — model not fitted")
}
```

This applies to **all 25 files** and should be fixed globally.

#### Unnecessary `Option<T>` wrapping

Parameters are stored as `Option<T>` but are **always** `Some(...)` after `fit()` — there is no legitimate `None` state post-construction. Consider either:
- Storing parameters as `T` directly (not wrapped in `Option`), or
- Returning `Option<&T>` from the getter and letting callers handle the `None` case without panicking.

A library API should not expose infallible-looking getters that can silently panic.

---

### 🟡 Moderate

#### No tests added

The PR adds 294 lines with zero new tests. Every getter should have at least a basic round-trip test: fit a model, call `.parameters()`, assert the returned values match what was passed in. Existing test modules in each file are the right place.

#### Breaking API changes not flagged

- In `decision_tree_classifier.rs`, `parameters()` was a private trait method and is now `pub fn` — this widens the public API and is a semver-breaking change.
- In `svc.rs`, `seed: Option<u64>` is changed from private to `pub seed` — this appears to be an unintentional side-effect and is also a public API change.

Both should be explicitly documented in the PR description and changelog, and the maintainers should assess the version bump implications.

---

### 🟢 Minor

- **Trailing commas**: Some struct field additions have trailing commas, others do not (e.g. `ExtraTreesRegressor`, `DecisionTreeRegressor`). Please run `cargo fmt` before merging.
- **`Option::None` / `Option::Some` style**: Several places use `Option::None` and `Option::Some(...)` instead of the idiomatic `None` / `Some(...)`. Not incorrect, but inconsistent with the rest of the codebase.
- **Doc comments**: All getters share the same generic boilerplate. Comments should mention the specific return type and note that the method will panic if called on an unfitted model instance.
- **Hidden `.clone()` cost**: Several `fit()` methods now clone fields before storing `parameters` (e.g. `parameters.alpha.clone()`, `parameters.distance.clone()`). This is necessary given the design, but worth acknowledging — especially for types like distance functions that may be non-trivially cloneable.

---

### Summary of requested changes

1. Replace all `assert! + unwrap()` with `.expect("...")`.
2. Re-evaluate `Option<T>` wrapping — store as `T` directly or return `Option<&T>`.
3. Add at least one test per getter verifying round-trip correctness.
4. Explicitly document and flag the breaking API changes (`DecisionTreeClassifier::parameters` visibility, `SVCParameters::seed` visibility).
5. Run `cargo fmt` to resolve trailing comma and style inconsistencies.

Mec-iS:
## ⛔ Build broken: serialisation (`serde`) failures

This PR **breaks compilation with `--all-features`** due to the new `parameters` fields added to model structs. The CI build (`cargo build --all-features`) fails with `E0277` errors because the parameter types being stored do not satisfy the `Default` trait bound required by serde's derived `Deserialize` implementation.

Confirmed failures on at least the following structs:

- `DBSCAN` / `DBSCANParameters<TX, D>` — `D: Default` not satisfied (`src/cluster/dbscan.rs:67`)
- `LogisticRegression` / `LogisticRegressionParameters<TX>` — `TX: Default` not satisfied (`src/linear/logistic_regression.rs:181`)
- `BernoulliNB` / `BernoulliNBParameters<TX>` — `TX: Default` not satisfied (`src/naive_bayes/bernoulli.rs:360`)
- `KNNClassifier` / `KNNClassifierParameters<TX, D>` — `D: Default` not satisfied (`src/neighbors/knn_classifier.rs:86`)

The root cause is that the derive macro for `Deserialize` requires all fields to implement `Default` (for the missing-field fallback path), but the type parameters on these structs (particularly distance functions `D` and float types `TX`) do not implement `Default`. Because the PR did not run the full test suite before submitting, these regressions went undetected.

### How to fix

For each affected struct, the options are:

1. **Add `#[serde(skip)]` to the `parameters` field** — this is the simplest fix if backward-compatible serialisation is acceptable (deserialized models will have `parameters: None`):
   ```rust
   #[cfg_attr(feature = "serde", serde(skip))]
   parameters: Option<DBSCANParameters<TX, D>>,
   ```
   Note: this means parameters are not round-tripped through serde, which partially undermines the stated goal of the PR.

2. **Implement `Deserialize` manually** for the affected structs — more work, but preserves full serde support.

3. **Add `+ Default` bounds to the type parameters** — as suggested by the compiler, but this is a **breaking change** to the public API and would ripple across many downstream users.

Option 1 is the pragmatic starting point but the trade-off must be documented clearly.

---

Please run the full test suite including the `serde` feature flag before pushing updates, as described in the [CONTRIBUTING guide](https://github.com/smartcorelib/smartcore/blob/development/CONTRIBUTING.md):

```sh
cargo test --all-features
cargo build --all-features
```

This PR should not be merged until these compilation errors are resolved.

Mec-iS:
Thanks @HubertMajewski
Let's me know if you want to proceed with the fixes yourself

HubertMajewski:
Hi @Mec-iS 

I initially committed them to unblock on a project I am working on. These changes are highly reasonable; I would be more than happy to apply these fixes.

Mec-iS:
Please run `cargo build --all-features` to check the build locally.

HubertMajewski:
@Mec-iS Please feel free to review now and let me know of any other additional changes.

Mec-iS:
The lint check is still failing. Please follow the commands as in CONTRIBUTING.md instructions, in particular run `cargo clippy ...`. you can read where the linting fails at https://github.com/smartcorelib/smartcore/actions/runs/29898188107/job/88903474713?pr=379

This is quite a massive PR with 1500+ lines of code so we need to be careful to add real value for the user relative to the complexity we add in the code.

Thanks again for this PR @HubertMajewski — exposing `.parameters()` consistently is a genuinely useful addition. After reviewing the source code in depth, here are the issues found, grouped by severity.

---

### 🟠 High Severity

**1. `SVC::parameters()` silently returns `None` after deserialization**

`SVC` stores its parameters as a lifetime-bound borrow:

```rust
parameters: Option<&'a SVCParameters<TX, TY, X, Y>>,
```

Because a reference cannot be serialized, the field is annotated with `#[serde(skip)]`, meaning any deserialized `SVC` will always return `None` from `.parameters()`. The test `test_returns_none_on_no_parameters` explicitly validates this degraded state rather than a successful round-trip — which is the opposite of what the PR description promises ("Added support for retrieving model hyperparameters after deserialization where applicable").

Compare this with `LogisticRegression`, which stores parameters by value (owned `Option<LogisticRegressionParameters<TX>>`), enabling a true serde round-trip that is correctly tested.

**Recommendation:** Either clone `SVCParameters` into the fitted `SVC` at fit time (taking ownership, matching the `LogisticRegression` pattern), or explicitly document in the `SVC::parameters()` doc comment that this method **always returns `None` after deserialization**, so callers are not silently misled.

---

**2. `MultiClassSVC` has no `.parameters()` getter**

`SVC` (binary) received a `.parameters()` getter, but `MultiClassSVC` — which wraps multiple `SVC` instances and accepts the same `SVCParameters` — did not. This leaves the API inconsistent: a user working with multi-class SVM cannot inspect the parameters used, defeating one of the stated goals of the PR.

**Recommendation:** Add a `parameters()` getter to `MultiClassSVC`, or add a code comment explicitly justifying the omission.

---

### 🟡 Medium Severity

**3. Custom `Deserialize` impl — correct, but needs a comment**

The hand-written `Deserialize` for `LogisticRegressionParameters` uses a private helper struct to avoid requiring a `Default` bound on the generic type `T`. The implementation is sound and the pattern is a recognised Serde idiom. However, there is no doc comment explaining *why* `#[derive(Deserialize)]` is insufficient here. Future maintainers may not understand why the custom impl exists and could accidentally replace it with a derive.

**Recommendation:** Add a short `// SERDE: manual impl avoids requiring Default on T; #[derive(Deserialize)] would add an implicit Default bound` comment above the impl block. Also verify that all other parameter structs changed in this PR apply the same strategy consistently (or use `#[serde(default)]` with an explicit `Default` impl where appropriate).

---

**4. Audit for missing models**

The `src/` tree contains: `cluster`, `decomposition`, `ensemble`, `linear` (5 models), `naive_bayes`, `neighbors`, `svm` (svc, svr), `tree`, and `xgboost`. The PR touches 25 files, but it is not immediately clear from the PR description whether `svr.rs`, the `naive_bayes` variants, `neighbors` (KNN), `decomposition` (PCA/SVD), and `xgboost` trees all received `.parameters()` getters. Any omissions will leave the API inconsistent.

**Recommendation:** Add a checklist to the PR description explicitly listing every model struct that was updated and any that were intentionally skipped (with justification).

HubertMajewski:
@Mec-iS thoughts?

Mec-iS:
it is a difficult call. we should know how/how much users will leverage serialisation. It is a lot of code lines for something that has limited usage, this is a use-case for macros probably; we may consider using macros only for serialisation.
One of our design principles is to not use macros but we could make an exception.

I’d favor a narrower serde change here: keep this PR limited to models that own their parameter structs and can round-trip cleanly after deserialization. SmartCore already has precedent for targeted serde work — v0.4.8 added serde support for XGRegressor as a focused change rather than a broad rewrite.
github

In the current code patterns, owned Option<Parameters> models are straightforward, while borrowed-parameter models like the SVM family introduce much more complexity and inconsistent post-deserialization behavior.
smartcore-v0410.md

That gives users value now, keeps the PR reviewable, and avoids making a macro exception unless we later decide the broader serialization problem is worth solving separately.
smartcore_prompt.txt

Mec-iS:
@HubertMajewski 

I think the safest way to move this PR forward is to narrow it to models that **own** their parameters and defer borrowed-parameter models to a follow-up.

### Proposed checklist

- [ ] Keep `.parameters()` only for fitted models that store `parameters: Option<...Parameters>` **by value** and can support a clean serde round-trip.
- [ ] Keep the current approach in owned-parameter files such as:
  - `src/tree/decision_tree_classifier.rs`
  - `src/tree/base_tree_regressor.rs`
- [ ] Keep other updated files in this PR only if they follow the same owned-parameters pattern as the tree models and do not require public API breakage to compile with `--all-features`.

- [ ] Remove or revert `.parameters()` work from borrowed-parameter SVM files in this PR:
  - `src/svm/svc.rs`
  - `src/svm/svr.rs`
- [ ] Remove or revert `MultiClassSVC` parameter-getter work from `src/svm/svc.rs`, since `MultiClassSVC` currently wraps classifiers rather than storing owned parameters directly.
- [ ] Remove tests that treat “`.parameters()` returns `None` after deserialization” as an acceptable success case for SVMs, because that does not really meet the stated feature goal.

- [ ] For owned generic parameter structs that still need manual serde support, keep the manual `Deserialize` implementation and add a short comment explaining that it avoids the implicit `Default` bound from `derive(Deserialize)`.
- [ ] Run `cargo build --all-features` and `cargo clippy --all-targets --all-features -- -D warnings` after the scope reduction.
- [ ] Update the PR description with an explicit “in scope / out of scope” list so reviewers can verify which models are intentionally included now and which are deferred.

This would give us a smaller, mergeable serde improvement now and leave the broader SVM/macros/ownership discussion for a dedicated follow-up.

Mec-iS:
I am closing this considering the relevant updates brought on by v0.6
Please let me know if you want to reopen.

PR #381, opened 2026-07-12
Add DenseMatrix constructor for owned ndarray matrices

### LINKED ISSUE #326: Implement a convenience method DenseMatrix::from_ndarray2 (in case the ndarray feature is on)
Example code (place in a module that depends on both smartcore and ndarray)

```rust
use std::fmt::{Debug, Display};
use ndarray::{ArrayBase, Ix2, OwnedRepr};
use smartcore::linalg::basic::arrays::Array2;
use smartcore::linalg::basic::matrix::DenseMatrix;

// Convert an ndarray owned 2D array into a smartcore DenseMatrix by copying.
pub fn from_ndarray2<T>(a: &ArrayBase<OwnedRepr<T>, Ix2>) -> DenseMatrix<T>
where
    T: Debug + Display + Copy + Sized,
{
    // SmartCore’s Array2::from_iterator(iter, nrows, ncols, axis)
    // axis == 0 means values are consumed in row-major and DenseMatrix will be created row-major=false internally,
    // which DenseMatrix::from_iterator translates correctly into its internal representation. [1]
    DenseMatrix::from_iterator(a.iter().copied(), a.nrows(), a.ncols(), 0)
}
```

Why this works

- smartcore defines a generic Array2::from_iterator for DenseMatrix that builds a matrix from any iterator of T with explicit shape and an axis flag to reconcile layout. Passing axis=0 matches the straightforward row-wise iteration over ndarray’s elements produced by a.iter().[1]
- DenseMatrix’s Array2::from_iterator is implemented as DenseMatrix::new(nrows, ncols, iter.collect(), axis != 0), and DenseMatrix::new enforces shape consistency; this path is the same used throughout the library for safe construction.[1]

If a borrowed view is needed

- For ArrayView<'_, T, Ix2> (or ArrayViewMut), convert to an owned array first: let owned = view.to_owned(); then call ndarray_to_densematrix(\&owned).[1]

Optional: return Result instead of panicking

```rust
pub fn from_ndarray2_result<T>(
    a: &ArrayBase<OwnedRepr<T>, Ix2>,
) -> Result<DenseMatrix<T>, smartcore::error::Failed>
where
    T: Debug + Display + Copy + Sized,
{
    // Use the same constructor but avoid unwrap; if you prefer, replicate logic with DenseMatrix::new.
    Ok(DenseMatrix::from_iterator(a.iter().copied(), a.nrows(), a.ncols(), 0))
}
```

Notes

- This copies data; a zero-copy adapter would require exposing DenseMatrix internals or constructing a DenseMatrix over ndarray’s buffer with exact column-major constraints. That’s not exposed in the current public API, so copying is the safe route.
- smartcore already offers ndarray bindings to operate on ndarray-native matrices directly by implementing Array2 for ndarray::Array2, so in many cases DenseMatrix isn’t necessary — algorithms accept types implementing the Array2 trait. Prefer passing ndarray arrays directly to smartcore algorithms when possible.

oiahoon:
Fixes #326

### Checklist

### Current behaviour

Converting an owned two-dimensional ndarray into a `DenseMatrix` requires callers to reproduce the iterator, shape, and axis conversion manually.

### New expected behaviour

When the `ndarray-bindings` feature is enabled, `DenseMatrix::from_ndarray2` copies an owned two-dimensional ndarray while preserving its logical row and column order. The regression test covers both a rectangular array and an owned array with transposed layout.

#### Added
- Add `DenseMatrix::from_ndarray2` for owned two-dimensional ndarrays.
- Add API documentation and layout regression coverage.

### Validation

- `cargo build --all-features`
- `cargo build --no-default-features`
- `cargo test --all-features`
- `cargo test --no-default-features --features ndarray-bindings`
- `cargo fmt --all -- --check`
- `cargo doc --all-features --no-deps`

The strict clippy command reaches one existing `clippy::for_kv_map` finding in `src/algorithm/neighbour/fastpair.rs:149` on Rust 1.97.0. The same finding occurs on the unchanged development branch; allowing only that baseline lint leaves the all-features clippy run clean.

### DIFF
--- a/src/linalg/ndarray/matrix.rs
+++ b/src/linalg/ndarray/matrix.rs
@@ -4,6 +4,7 @@ use std::ops::Range;
 use crate::linalg::basic::arrays::{
     Array as BaseArray, Array2, ArrayView1, ArrayView2, MutArray, MutArrayView2,
 };
+use crate::linalg::basic::matrix::DenseMatrix;
 
 use crate::linalg::traits::cholesky::CholeskyDecomposable;
 use crate::linalg::traits::evd::EVDDecomposable;
@@ -18,6 +19,46 @@ use ndarray::{s, Array, ArrayBase, ArrayView, ArrayViewMut, Axis, Ix2, Order, Ow
 // ---------------------------------------------------------------------------
 // ArrayBase<OwnedRepr<T>, Ix2>  (owned 2-D array)
 // ---------------------------------------------------------------------------
+
+const ROW_MAJOR_AXIS: u8 = 0;
+
+impl<T: Debug + Display + Copy> DenseMatrix<T> {
+    /// Copies an owned two-dimensional ndarray into a [`DenseMatrix`].
+    ///
+    /// The resulting matrix uses row-major (C) storage regardless of the
+    /// memory layout of the source array.
+    ///
+    /// # Notes
+    ///
+    /// [`ndarray::Array2::iter`] always yields elements in logical row-major
+    /// order, independent of whether the source is C- or Fortran-ordered. This
+    /// invariant makes transposed-layout conversion correct.
+    ///
+    /// # Panics
+    ///
+    /// Panics if `nrows * ncols` overflows `usize`. An empty array (zero
+    /// rows or zero columns) does not panic.
+    ///
+    /// # Examples
+    ///
+    /// ```
+    /// use ndarray::Array2;
+    /// use smartcore::linalg::basic::arrays::Array;
+    /// use smartcore::linalg::basic::matrix::DenseMatrix;
+    ///
+    /// let array = Array2::from_shape_vec(
+    ///     (3, 4),
+    ///     (0..12).map(|value| value as f64).collect(),
+    /// ).unwrap();
+    /// let matrix = DenseMatrix::from_ndarray2(&array);
+    /// assert_eq!(matrix.shape(), (3, 4));
+    /// assert_eq!(*matrix.get((1, 2)), 6.0);
+    /// ```
+    pub fn from_ndarray2(a: &ndarray::Array2<T>) -> Self {
+        // iter() yields logical row-major order regardless of memory layout.
+        Self::from_iterator(a.iter().copied(), a.nrows(), a.ncols(), ROW_MAJOR_AXIS)
+    }
+}
 
 impl<T: Debug + Display + Copy + Sized> BaseArray<T, (usize, usize)>
     for ArrayBase<OwnedRepr<T>, Ix2>
@@ -223,6 +264,53 @@ impl<T: Debug + Display + Copy + Sized> MutArrayView2<T> for ArrayViewMut<'_, T,
 mod tests {
     use super::*;
     use ndarray::arr2;
+
+    #[test]
+    fn test_dense_matrix_from_ndarray2() {
+        let input = arr2(&[[1, 2, 3], [4, 5, 6]]);
+        let matrix = DenseMatrix::from_ndarray2(&input);
+        let expected = DenseMatrix::from_2d_array(&[&[1, 2, 3], &[4, 5, 6]]).unwrap();
+        assert_eq!(matrix, expected);
+
+        let transposed = input.reversed_axes();
+        let matrix = DenseMatrix::from_ndarray2(&transposed);
+        let expected = DenseMatrix::from_2d_array(&[&[1, 4], &[2, 5], &[3, 6]]).unwrap();
+        assert_eq!(matrix, expected);
+    }
+
+    #[test]
+    fn test_dense_matrix_from_ndarray2_square() {
+        let input = arr2(&[[1, 2], [3, 4]]);
+        let matrix = DenseMatrix::from_ndarray2(&input);
+        let expected = DenseMatrix::from_2d_array(&[&[1, 2], &[3, 4]]).unwrap();
+        assert_eq!(matrix, expected);
+    }
+
+    #[test]
+    fn test_dense_matrix_from_ndarray2_row_vector() {
+        let input = arr2(&[[10, 20, 30, 40]]);
+        let matrix = DenseMatrix::from_ndarray2(&input);
+        let expected = DenseMatrix::from_2d_array(&[&[10, 20, 30, 40]]).unwrap();
+        assert_eq!(matrix, expected);
+        assert_eq!(matrix.shape(), (1, 4));
+    }
+
+    #[test]
+    fn test_dense_matrix_from_ndarray2_col_vector() {
+        let input = arr2(&[[10], [20], [30], [40]]);
+        let matrix = DenseMatrix::from_ndarray2(&input);
+        let expected = DenseMatrix::from_2d_array(&[&[10], &[20], &[30], &[40]]).unwrap();
+        assert_eq!(matrix, expected);
+        assert_eq!(matrix.shape(), (4, 1));
+    }
+
+    #[test]
+    fn test_dense_matrix_from_ndarray2_empty() {
+        let input = ndarray::Array2::<i32>::zeros((0, 0));
+        let matrix = DenseMatrix::from_ndarray2(&input);
+        assert!(matrix.is_empty());
+        assert_eq!(matrix.shape(), (0, 0));
+    }
 
     #[test]
     fn test_get_row() {


Mec-iS:
Thanks for this! I have started the CI and I will provide a code review asap.

Mec-iS (changes requested):
## Code Review: `DenseMatrix::from_ndarray2`

Thank you for this addition — the feature fills a real gap and the transposed-layout regression test is a great catch. After reviewing the diff against the `DenseMatrix` internals I have **5 actionable suggestions** before merging, detailed as inline comments below.

### Summary

| # | Area | Severity |
|---|---|---|
| 1 | API: use `Array2<T>` alias instead of `ArrayBase<OwnedRepr<T>, Ix2>` | Minor |
| 2 | Trait bounds: redundant `Sized` | Nit |
| 3 | Magic `0` axis literal unexplained | Minor |
| 4 | Doc-comment missing `# Panics` and `# Notes` (layout invariant) | Minor |
| 5 | Test coverage missing edge cases | Minor |

### Suggested replacement for the entire new block

```rust
// `axis = 0` passed to DenseMatrix::from_iterator means row-major storage
// (column_major = false).  Named here to avoid a magic literal.
const ROW_MAJOR_AXIS: u8 = 0;

impl<T: Debug + Display + Copy> DenseMatrix<T> {
    /// Copies an owned two-dimensional ndarray into a [`DenseMatrix`].
    ///
    /// The resulting matrix uses row-major (C) storage regardless of the
    /// memory layout of the source array.
    ///
    /// # Notes
    ///
    /// [`ndarray::Array2::iter`] always yields elements in **logical
    /// row-major order** (row 0 col 0, row 0 col 1, …, row n col m),
    /// independent of whether the source array is stored in C or Fortran
    /// (column-major) order.  This is what makes the transposed-axis test
    /// pass correctly and should be preserved by future refactors.
    ///
    /// # Panics
    ///
    /// Panics if `nrows * ncols` overflows `usize`, which is unreachable on
    /// 32- and 64-bit targets in practice.  An empty (0 rows or 0 cols)
    /// array produces an empty `DenseMatrix` without panicking.
    ///
    /// # Examples
    ///
    /// ```
    /// use ndarray::Array2;
    /// use smartcore::linalg::basic::matrix::DenseMatrix;
    ///
    /// // Build a 3×4 ndarray from a flat vector and convert it.
    /// let nd: Array2<f64> = Array2::from_shape_vec(
    ///     (3, 4),
    ///     (0..12).map(|x| x as f64).collect(),
    /// ).unwrap();
    /// let dm = DenseMatrix::from_ndarray2(&nd);
    /// assert_eq!(dm.shape(), (3, 4));
    /// assert_eq!(*dm.get((1, 2)), 6.0);
    /// ```
    pub fn from_ndarray2(a: &ndarray::Array2<T>) -> Self {
        // .iter() yields elements in logical row-major order regardless of
        // the underlying memory layout, so ROW_MAJOR_AXIS is always correct.
        Self::from_iterator(a.iter().copied(), a.nrows(), a.ncols(), ROW_MAJOR_AXIS)
    }
}
```

### Suggested additional tests

```rust
#[test]
fn test_dense_matrix_from_ndarray2_square() {
    let input = arr2(&[[1, 2], [3, 4]]);
    let matrix = DenseMatrix::from_ndarray2(&input);
    let expected = DenseMatrix::from_2d_array(&[&[1, 2], &[3, 4]]).unwrap();
    assert_eq!(matrix, expected);
}

#[test]
fn test_dense_matrix_from_ndarray2_row_vector() {
    let input = arr2(&[[10, 20, 30, 40]]);
    let matrix = DenseMatrix::from_ndarray2(&input);
    let expected = DenseMatrix::from_2d_array(&[&[10, 20, 30, 40]]).unwrap();
    assert_eq!(matrix, expected);
    assert_eq!(matrix.shape(), (1, 4));
}

#[test]
fn test_dense_matrix_from_ndarray2_col_vector() {
    let input = arr2(&[[10], [20], [30], [40]]);
    let matrix = DenseMatrix::from_ndarray2(&input);
    let expected = DenseMatrix::from_2d_array(&[&[10], &[20], &[30], &[40]]).unwrap();
    assert_eq!(matrix, expected);
    assert_eq!(matrix.shape(), (4, 1));
}

#[test]
fn test_dense_matrix_from_ndarray2_empty() {
    // 0×0 empty array — must not panic
    let input = ndarray::Array2::<i32>::zeros((0, 0));
    let matrix = DenseMatrix::from_ndarray2(&input);
    assert!(matrix.is_empty());
    assert_eq!(matrix.shape(), (0, 0));
}
```

All other existing tests and logic in the file are correct and should remain untouched.

Mec-iS (commented):
Inline comment batch — see individual comments below for each suggestion location.

Mec-iS:
please check the linting, see all the commands to run in CONTRIBUTING.md

src/linalg/ndarray/matrix.rs
```diff
@@ -19,6 +20,26 @@ use ndarray::{s, Array, ArrayBase, ArrayView, ArrayViewMut, Axis, Ix2, Order, Ow
 // ArrayBase<OwnedRepr<T>, Ix2>  (owned 2-D array)
 // ---------------------------------------------------------------------------
 
+impl<T: Debug + Display + Copy + Sized> DenseMatrix<T> {
```
Mec-iS: **[Suggestion 1 — Redundant `Sized` bound]**

`Sized` is implicit on every generic type parameter in Rust and should be removed to reduce noise:

```rust
// before
impl<T: Debug + Display + Copy + Sized> DenseMatrix<T> {

// after
impl<T: Debug + Display + Copy> DenseMatrix<T> {
```

src/linalg/ndarray/matrix.rs
```diff
@@ -19,6 +20,26 @@ use ndarray::{s, Array, ArrayBase, ArrayView, ArrayViewMut, Axis, Ix2, Order, Ow
...
+    /// Copies an owned two-dimensional ndarray into a `DenseMatrix`.
```
Mec-iS: **[Suggestion 2 — Doc-comment: missing `# Panics` and `# Notes`]**

The current doc-comment only has `# Examples`. Please add:
- A `# Notes` section documenting the critical layout invariant: `ndarray::iter()` always yields elements in **logical row-major order** regardless of whether the source array is C- or Fortran-ordered. This is what makes the transposed-axis test pass and must not be silently removed in future refactors.
- A `# Panics` section: panics if `nrows * ncols` overflows `usize` (unreachable on 32/64-bit targets). An empty (0 rows or 0 cols) array does not panic.

Also, consider replacing the `arr2` example with a slightly more realistic use-case (e.g. `Array2::from_shape_vec`) to differentiate it from the unit test below.

```rust
/// Copies an owned two-dimensional ndarray into a [`DenseMatrix`].
///
/// The resulting matrix uses row-major (C) storage regardless of the
/// memory layout of the source array.
///
/// # Notes
///
/// [`ndarray::Array2::iter`] always yields elements in **logical
/// row-major order** (row 0 col 0, row 0 col 1, …), independent of
/// whether the source is C- or Fortran-ordered. This is the invariant
/// that makes transposed-layout conversion correct here.
///
/// # Panics
///
/// Panics if `nrows * ncols` overflows `usize` (unreachable on 32/64-bit
/// targets). An empty (0 rows or 0 cols) array does not panic.
///
/// # Examples
///
/// ```
/// use ndarray::Array2;
/// use smartcore::linalg::basic::matrix::DenseMatrix;
///
/// let nd: Array2<f64> = Array2::from_shape_vec(
///     (3, 4),
///     (0..12).map(|x| x as f64).collect(),
/// ).unwrap();
/// let dm = DenseMatrix::from_ndarray2(&nd);
/// assert_eq!(dm.shape(), (3, 4));
/// assert_eq!(*dm.get((1, 2)), 6.0);
/// ```
```

src/linalg/ndarray/matrix.rs
```diff
@@ -19,6 +20,26 @@ use ndarray::{s, Array, ArrayBase, ArrayView, ArrayViewMut, Axis, Ix2, Order, Ow
...
+    ///
+    /// let array = arr2(&[[1, 2, 3], [4, 5, 6]]);
+    /// let matrix = DenseMatrix::from_ndarray2(&array);
+    /// let expected = DenseMatrix::from_2d_array(&[&[1, 2, 3], &[4, 5, 6]]).unwrap();
+    ///
+    /// assert_eq!(matrix, expected);
+    /// ```
+    pub fn from_ndarray2(a: &ArrayBase<OwnedRepr<T>, Ix2>) -> Self {
+        Self::from_iterator(a.iter().copied(), a.nrows(), a.ncols(), 0)
+    }
```
Mec-iS, line 60: **[Suggestion 3 — API signature: prefer `Array2<T>` alias]**

The public signature exposes `ArrayBase<OwnedRepr<T>, Ix2>` which leaks ndarray’s internal generic machinery. The canonical public alias `ndarray::Array2<T>` is cleaner and is what users will have in scope:

```rust
// before
pub fn from_ndarray2(a: &ArrayBase<OwnedRepr<T>, Ix2>) -> Self {

// after
pub fn from_ndarray2(a: &ndarray::Array2<T>) -> Self {
```

Note: `ndarray::Array2<T>` is simply a type alias for `ArrayBase<OwnedRepr<T>, Ix2>`, so this is a zero-cost, non-breaking change.

src/linalg/ndarray/matrix.rs
```diff
@@ -19,6 +20,26 @@ use ndarray::{s, Array, ArrayBase, ArrayView, ArrayViewMut, Axis, Ix2, Order, Ow
...
+}
```
Mec-iS, line 61: **[Suggestion 4 — Magic `0` axis literal]**

The `0` here is opaque without context. After checking the entire codebase, `axis = 0` means row-major storage (`column_major = false`) in `DenseMatrix::from_iterator` — it is the universal convention used in `dbscan.rs`, `csv.rs`, `logistic_regression.rs` and elsewhere.

Please introduce a named constant at the top of the `impl` block (or file scope) and add an inline comment:

```rust
// At file/module scope:
const ROW_MAJOR_AXIS: u8 = 0;

// In the method body:
pub fn from_ndarray2(a: &ndarray::Array2<T>) -> Self {
    // .iter() yields logical row-major order regardless of memory layout,
    // so ROW_MAJOR_AXIS (= 0, i.e. column_major=false) is always correct.
    Self::from_iterator(a.iter().copied(), a.nrows(), a.ncols(), ROW_MAJOR_AXIS)
}
```

src/linalg/ndarray/matrix.rs
```diff
@@ -224,6 +245,19 @@ mod tests {
...
+    fn test_dense_matrix_from_ndarray2() {
+        let input = arr2(&[[1, 2, 3], [4, 5, 6]]);
+        let matrix = DenseMatrix::from_ndarray2(&input);
+        let expected = DenseMatrix::from_2d_array(&[&[1, 2, 3], &[4, 5, 6]]).unwrap();
+        assert_eq!(matrix, expected);
+
+        let transposed = input.reversed_axes();
+        let matrix = DenseMatrix::from_ndarray2(&transposed);
+        let expected = DenseMatrix::from_2d_array(&[&[1, 4], &[2, 5], &[3, 6]]).unwrap();
+        assert_eq!(matrix, expected);
```
Mec-iS, line 278: **[Suggestion 5 — Missing edge-case tests]**

The current test covers a rectangular array and a transposed (Fortran-order) layout — both good. The following edge cases are missing:

```rust
#[test]
fn test_dense_matrix_from_ndarray2_square() {
    // Square matrix (nrows == ncols)
    let input = arr2(&[[1, 2], [3, 4]]);
    let matrix = DenseMatrix::from_ndarray2(&input);
    let expected = DenseMatrix::from_2d_array(&[&[1, 2], &[3, 4]]).unwrap();
    assert_eq!(matrix, expected);
}

#[test]
fn test_dense_matrix_from_ndarray2_row_vector() {
    // 1×N degenerate shape
    let input = arr2(&[[10, 20, 30, 40]]);
    let matrix = DenseMatrix::from_ndarray2(&input);
    let expected = DenseMatrix::from_2d_array(&[&[10, 20, 30, 40]]).unwrap();
    assert_eq!(matrix, expected);
    assert_eq!(matrix.shape(), (1, 4));
}

#[test]
fn test_dense_matrix_from_ndarray2_col_vector() {
    // N×1 degenerate shape
    let input = arr2(&[[10], [20], [30], [40]]);
    let matrix = DenseMatrix::from_ndarray2(&input);
    let expected = DenseMatrix::from_2d_array(&[&[10], &[20], &[30], &[40]]).unwrap();
    assert_eq!(matrix, expected);
    assert_eq!(matrix.shape(), (4, 1));
}

#[test]
fn test_dense_matrix_from_ndarray2_empty() {
    // 0×0 empty array — must not panic
    let input = ndarray::Array2::<i32>::zeros((0, 0));
    let matrix = DenseMatrix::from_ndarray2(&input);
    assert!(matrix.is_empty());
    assert_eq!(matrix.shape(), (0, 0));
}
```

Mec-iS (commented):
Posting 5 targeted inline comments (line-level) to accompany the earlier review. Each comment maps directly to one of the 5 suggestions in the REQUEST_CHANGES review above.

src/linalg/ndarray/matrix.rs
oiahoon: Addressed in d4bc985. The redundant Sized bound was removed from the constructor impl.

src/linalg/ndarray/matrix.rs
oiahoon: Addressed in d4bc985. The row-major axis is now a named constant with an explanatory iterator comment.

src/linalg/ndarray/matrix.rs
oiahoon: Addressed in d4bc985. The public parameter now uses ndarray::Array2<T>.

Mec-iS:
good! please @oiahoon check the linting failure in the CI and confirm that the suggested tests have been added or motivate why they should not be added.

Mec-iS:
will go in v0.5.4

PR #387, opened 2026-08-07
docs: document Cargo.toml features

### LINKED ISSUE #68: Document features in Cargo.toml
We should add a description to each feature defined in smartcore in the Cargo.toml file, we could use as example the [Cargo.toml file of the serde project](https://github.com/serde-rs/serde/blob/d7ccef0cac8b3703eee47edee19f6bce0109af27/serde/Cargo.toml#L26).

We should add also the documentation for the "serde" feature that currently is not present in the Cargo.toml file.

Not sure if we can use [this](https://github.com/rust-lang/api-guidelines/discussions/43#discussioncomment-233405).

teddytennant:
## Problem
Optional features in `Cargo.toml` had little or no description, so users had to dig through source or README to learn what each flag enables. Issue #68 asked for serde-style feature documentation, including the missing `serde` feature note.

## Change
Document every feature in `Cargo.toml` with short comments in the style of [serde's Cargo.toml](https://github.com/serde-rs/serde/blob/master/serde/Cargo.toml):

- `default` - empty (WASM-friendly defaults)
- `serde` - Serialize/Deserialize for models; enables typetag on non-wasm
- `ndarray-bindings` - optional ndarray DenseMatrix bindings
- `datasets` - sample datasets and generators (pulls in `std_rand` + `serde`)
- `std_rand` - StdRng / thread_rng instead of SmallRng
- `js` - getrandom wasm_js backend for browser builds

Feature names were checked against `cfg(feature = ...)` usage in the crate (`serde`, `datasets`, `ndarray-bindings`, `std_rand`, `js`).

## Test plan
- [x] `cargo metadata --no-deps` parses features successfully
- [x] Confirmed feature names match cfg gates in source
- Docs-only change; no code behavior change

Fixes #68

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -33,13 +33,27 @@ ordered-float = "5.1.0"
 [target.'cfg(not(target_arch = "wasm32"))'.dependencies]
 typetag = { version = "0.2", optional = true }
 
+### FEATURES #################################################################
+
 [features]
 default = []
+
+# Enable serde Serialize/Deserialize for models and related types.
+# Also enables typetag on non-wasm targets for trait object serialization.
 serde = ["dep:serde", "dep:typetag"]
+
+# Optional bindings for the ndarray crate (DenseMatrix <-> ndarray).
 ndarray-bindings = ["dep:ndarray"]
+
+# Sample datasets (iris, boston, digits, ...) and dataset generators.
+# Enables std_rand and serde because datasets ship serialized bundles.
 datasets = ["std_rand", "serde"]
+
+# Use the standard library RNG (StdRng / thread_rng) instead of SmallRng.
+# Required for non-deterministic seeding without an explicit seed.
 std_rand = ["rand/std_rng", "rand/std", "rand/thread_rng"]
-# used by wasm32-unknown-unknown for in-browser usage
+
+# Enable getrandom's wasm_js backend for wasm32-unknown-unknown (in-browser).
 js = ["getrandom/wasm_js"]
 
 [target.'cfg(target_arch = "wasm32")'.dependencies]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -33,9 +33,9 @@ ordered-float = "5.1.0"
 [target.'cfg(not(target_arch = "wasm32"))'.dependencies]
 typetag = { version = "0.2", optional = true }
 
-### FEATURES #################################################################
-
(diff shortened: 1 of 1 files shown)


Mec-iS:
## Overall Assessment

The PR is **well-scoped and correct** — it directly addresses issue #68 with no behavioral changes. The diff is minimal and the added comments are accurate and useful. It's ready to merge with one minor suggestion.

## What's Good

- **Accurate descriptions**: Each comment correctly reflects what the feature does. For example, noting that `datasets` pulls in `std_rand` and `serde` because datasets ship serialized bundles is genuinely useful context that wasn't there before .
- **Consistent style**: The serde-style inline comments above each feature (`# Enable ...`) are idiomatic for Rust crates and match the community convention .
- **The `### FEATURES ###` banner** adds visual scannability in the raw `Cargo.toml` file, which is a nice touch for maintainers.
- **The `js` feature comment** is an improvement over the old one-liner — it now clarifies *which* wasm target and *why* (in-browser getrandom backend) .

## Minor Suggestions

A few small improvements worth considering before merging:

1. **`default = []` lacks a comment.** Since the PR documents all other features, it would be consistent to add a comment like `# No features enabled by default (keeps the crate WASM-friendly and dependency-light)`. The PR description mentions this intent but it's not reflected in the actual file .

2. **`ndarray-bindings` description could be slightly more specific.** "DenseMatrix <-> ndarray" is good, but noting the directionality or that it's an *optional* interop layer (i.e., needed only when mixing ndarray types with smartcore APIs) would help users decide whether they need it.

3. **Coverage drop is unrelated but worth noting**: Codecov reports a drop from 45.59% → 44.00% coverage . This is a docs-only PR adding 2 files and 21 lines — the drop is almost certainly from the base `development` branch being 109 commits ahead, not from this change. Worth flagging to maintainers so they're not confused.

## Suggested Review Comment

You could add an inline suggestion on the `default = []` line:

```toml
# No features enabled by default; keeps the build WASM-compatible
# and avoids pulling in serde/rand unless explicitly requested.
default = []
```

The PR is otherwise **approved** — it's a low-risk, high-value documentation improvement that will directly benefit new contributors trying to understand feature flags without reading source code.

Mec-iS (dismissed):
thanks!

Mec-iS (approved):

Issue #322, opened 2025-07-13
Trait bounds of RandomForest::fit conflict with those of predict and other metrics


DaGaMs:
I am trying to perform cross validation of a RandomForest like this:

```rust
let cv_score = cross_validate(
    RandomForestClassifier::new(),
    x_train,
    y_train,
    RandomForestClassifierParameters::default()
        .with_criterion(SplitCriterion::ClassificationError)
        .with_n_trees(*n_tree)
        .with_m(*m_feat)
        .with_min_samples_split(*m_split)
        .with_min_samples_leaf(*m_leaf),
    &KFold::default().with_n_splits(5),
    &precision
).unwrap();
```

x_train is a `DenseMatrix<f32>` and y_train is a `Vec<u16>` of 0 and 1.

The problem is that `RandomForestClassifier::fit` expects `y` to be `Number + Ord`, whereas `precision` expects y to be `Number + RealNumber + FloatNumber`. As far as I can see, `RealNumber` can never be `Ord`, so `precision`, `f1` and `roc_auc_score` cannot be used in `cross_validate` with `RandomForectClassifier` directly.

For now, I worked around it by defining my own precision function that converts u16 to f32 before calling `precision`, but I suppose this should be fixed in the framework. The logical thing to do IMO would be to stop requiring `Ord` for `y` in `RandomForrestClassifier::fit`?

Mec-iS:
the initial point of view for this was that classification labels need to be integers otherwise which element belongs to which class is ambiguous. Integers have also the advantage to be ordered.

If you think that it is necessary to have labels with reals, you can propose your implementation of another classifier like `RandomForestContinuous` that implements your solution.

@DanielLacina please advice as you worked on trees recently

DaGaMs:
I agree that labels should be ordinal, but for some reason most of the metrics expect them to be reals!

In other words, I think the arguments to `precision`, `f1` etc need to have their trait bounds changed.

Mec-iS:
I understand but this would be a breaking change for existing programs so we need to think through if it is really necessary and eventually how to improve the situation

DaGaMs:
I see that - I just don't know how anyone would use `cross_validate` with RFs in the current situation 🤷‍♂️ 

I suppose the problem is that precision etc are meant to also work with regression algorithms, and that's why they want real numbers?

Mec-iS:
[...]

Issue #395, opened 2026-08-09
Stage 4: Metrics & distance coverage (tracking #391)


Mec-iS:
Tracking issue: #391 — Improve test coverage significantly (staged plan).

## Scope

Most metric files currently have only 1–2 tests. Broaden metric and distance coverage.

## Tasks

- [ ] `src/metrics/`: for `accuracy`, `precision`, `recall`, `f1`, `auc`, `r2`, `mean_absolute_error`, `mean_squared_error`:
  - perfect-score case (predictions == truth)
  - worst-score case
  - empty / single-sample input where defined
  - multiclass variants (extend `precision_multiclass` / `recall_multiclass` pattern)
- [ ] `src/metrics/cluster_hcv.rs` + `cluster_helpers.rs`: contingency matrix edge cases (single cluster, all-same-labels).
- [ ] `src/metrics/distance/`:
  - `euclidian`, `manhattan`, `minkowski`: zero vector, identical points, `p → ∞` limit for Minkowski, `p < 0` (already partly covered for negative p — extend).
  - `hamming`: identical vectors, all-different vectors, unequal-length handling.
  - `mahalanobis`: singular covariance matrix behavior, identity covariance reduces to Euclidean.
- [ ] `PairwiseDistance` struct: shape of output for non-square inputs, symmetry of pairwise matrix.

## Verification

- `cargo test --all-features`
- `cargo fmt --all -- --check`
- `cargo clippy --all-features -- -Drust-2018-idioms -Dwarnings`

## Do you want to work on this issue?
yes — PR will target `development`.

PR #438, opened 2026-08-21
fix(metrics): accept integer labels in precision/recall/f1 (#322)

Mec-iS:
Fixes #322

### Checklist

### Current behaviour
`RandomForestClassifier::fit` requires `TY: Number + Ord`, so classification labels are typically integers (`u16`, `i32`, ...). The classification metrics `precision`, `recall` and `f1` require `T: Number + RealNumber + FloatNumber`. No type satisfies both (`RealNumber` is float-only, floats are not `Ord`), so the same `y` cannot be passed to `fit` and to the metrics. In particular `cross_validate(RandomForestClassifier::new(), x, y, params, &KFold::default(), &precision)` does not compile with integer labels; users must write ad-hoc wrappers that convert labels to `f32`/`f64`.

### New expected behaviour
`Precision`, `Recall`, `F1` (structs and free functions) now require only `T: Number`. These metrics need only label equality and a canonical class key, never float arithmetic on labels, so the bound is relaxed rather than `Ord` being dropped from `fit` — integer labels stay the canonical representation for tree classifiers, as agreed in #322.

Implementation notes:
- `metrics/confusion.rs`: new `pub(crate) fn label_bits<T: Number>(v: T) -> u64` derives class keys via `to_f64` widening instead of `RealNumber::to_f64_bits` raw transmutation. `ConfusionCounts::new` is generic over `Number`.
- Float-input scores are bit-for-bit unchanged: for `f64` the conversion is the identity; for `f32` it is exact widening. `-0.0` / `+0.0` remain distinct classes. Integer labels convert exactly up to 2^53 in magnitude (documented).
- The `ClassificationMetrics<T>` facade keeps its float bounds because of `roc_auc_score`; integer users go through the free functions or the metric structs directly.

#### Fixed
- `precision`, `recall`, `f1` accept ordered integer labels (#322); `cross_validate` + `RandomForestClassifier` + `&precision` compiles with e.g. `Vec<u16>` targets.

#### Changed
- Patch version bump 0.6.8 → 0.6.9.

### Tests
- Integer-label unit tests for `precision` / `recall` / `f1` (binary + multiclass macro-average, known-answer values mirrored from the float cases).
- Regression test `cross_validate_with_precision_and_integer_labels` in `ensemble/random_forest_classifier.rs` reproducing the exact #322 scenario (`DenseMatrix<f32>`, `Vec<u16>`, 5-fold CV with `&precision`).
- `cargo test --all-features` green (501 lib tests + doctests), `cargo fmt --all -- --check` clean, `cargo clippy --all-features -- -Drust-2018-idioms -Drust-2024-compatibility -Dwarnings` clean.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,10 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [0.6.9]
+### Fixed
+- `metrics`: `precision`, `recall`, and `f1` (the free functions and the `Precision` / `Recall` / `F1` metric structs) now accept any label type that implements `Number`, including ordered integers such as `u16` or `i32`; labels no longer need to implement `RealNumber` (#322). The same integer labels can now feed `RandomForestClassifier::fit` and classification metrics inside `model_selection::cross_validate`. Class keys are derived through a shared `f64` conversion instead of raw float bit transmutation; scores for float inputs are unchanged.
+
 ## [0.6.8]
 ### Fixed
 - `linear/linear_regression.rs`: `LinearRegression::fit` / `fit_matrix` now return `Err(Failed::fit(...))` instead of panicking when the intercept-augmented system is underdetermined, i.e. `n_features + 1 > n_samples` (#435). Both the default SVD solver and the QR solver are covered.
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -2,7 +2,7 @@
 name = "smartcore"
 description = "Machine Learning in Rust."
 homepage = "https://smartcorelib.github.io/"
-version = "0.6.8"
+version = "0.6.9"
 authors = ["smartcore Developers"]
 edition = "2024"
 rust-version = "1.85"
--- a/src/ensemble/random_forest_classifier.rs
+++ b/src/ensemble/random_forest_classifier.rs
@@ -693,6 +693,65 @@ mod tests {
         assert!(accuracy(&y, &classifier.predict(&x).unwrap()) >= 0.95);
     }
 
+    // Regression test for #322: `cross_validate` with a `RandomForestClassifier`
+    // and the `precision` metric must accept the same ordered integer labels
+    // (`Number + Ord`) that `fit` accepts.
+    #[cfg_attr(
+        all(target_arch = "wasm32", not(target_os = "wasi")),
+        wasm_bindgen_test::wasm_bindgen_test
+    )]
+    #[test]
+    fn cross_validate_with_precision_and_integer_labels() {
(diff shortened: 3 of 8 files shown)


Mec-iS (commented):
## Review — looks good to merge, three non-blocking notes

Read the full diff of `fix/322-metric-trait-bounds` against `development` and ran independent verification beyond the CI suite.

### What was verified

**Correctness of the bound relaxation.** `Precision`/`Recall`/`F1` internals use only label equality (`Num`), `T::one()`, and a canonical class key — no float arithmetic on labels. `T: Number` is the tightest sound bound, and it strictly widens the old API, so the change is backward compatible. `ClassificationMetrics<T>` facade untouched (still float-gated by `roc_auc_score`); integer users go through the free functions, which the doc comments now say.

**`label_bits` semantics** (probed outside the test suite):
- `f32 → f64` widening is exact and injective — class grouping is preserved for all float inputs.
- `-0.0` / `+0.0` remain distinct classes; NaN payloads stay distinct (same convention as before).
- `1u16` and `1.0f64` map to the same key, so the binary positive-class lookup (`label_bits(T::one())`) works for integer labels.
- Integer labels are exact up to `2^53`; collisions only above that, as documented in the helper doc comment.

**Differential check (20k randomized trials, sorted-key deterministic sums):** old keying vs new keying gives bit-identical precision/recall/f1 for `f64` inputs and for non-negative `f32` inputs — max |delta| = 0. (A first unsorted probe showed ±1 ULP noise, but that comes from `HashMap` `RandomState` iteration order, which the released code already had — multiclass macro-averages sum over `HashMap::values()` both before and after this PR. Not introduced or worsened here.)

**Gates:** `cargo test --all-features` (501 lib tests + doctests, incl. the new #322 regression test), `fmt --check`, `clippy --all-features -- -Drust-2018-idioms -Drust-2024-compatibility -Dwarnings` — all clean.

### Notes (non-blocking)

1. **`label_bits` uses `.unwrap()` on `to_f64()`** (`confusion.rs:24`). Safe for all 11 primitive `Number` impls (their `to_f64` never returns `None`), but `Number` is implementable downstream; a custom type with a failing `to_f64` would panic with an opaque message. Consider `.expect("class label must convert to f64")`. Matches existing codebase style either way.
2. **CHANGELOG wording**: "scores for float inputs are unchanged" is exactly true for binary scores and deterministic runs; multiclass macro-averages were already nondeterministic at the 1-ULP level due to `HashMap` iteration order (pre-existing). Fine as written — just flagging the subtlety.
3. **Regression test uses `DenseMatrix<f64>`** while #322 reported `f32`. The conflict was on the label type (`Vec<u16>`), which the test does reproduce, so coverage is valid; mirroring the issue's `f32` features would be even closer. Nit.

### Out of scope, observed while reading
`src/metrics/tests_edge_cases.rs` (added in #395) is not declared in `metrics/mod.rs` and never compiles. Not this PR's job, but worth a tracking issue — its `Precision::<i32>` cases become valid after this PR lands.

PR #445, opened 2026-08-24
fix(xgboost): keep one row when subsample truncates the sample to zero (#444)

### LINKED ISSUE #444: XGRegressor::fit panics (subtract with overflow) when subsample truncates the sampled rows to zero
`XGRegressor::fit` panics with "attempt to subtract with overflow" when `subsample` is small enough that the sampled index set comes out empty.

`sample_without_replacement` sizes the sample as `(population_size as f64 * subsample_ratio) as usize`, which floors to 0 on small inputs. Three rows at `subsample = 0.3` gives 0, and a single row gives 0 for any `subsample < 1.0`. `TreeRegressor::fit` then calls `find_best_split` with an empty slice, and the loop bound `0..sorted_idxs.len() - 1` underflows.

The parameter guard in `fit` only rejects `subsample > 1.0` and `subsample <= 0.0`, so this input gets through.

Repro:

```rust
let x = DenseMatrix::from_2d_array(&[&[1.0, 2.0], &[3.0, 4.0], &[5.0, 6.0]]).unwrap();
let y = vec![1.0, 2.0, 3.0];
let params = XGRegressorParameters::default().with_subsample(0.3);
let _ = XGRegressor::fit(&x, &y, params);
```

Observed:

```
thread 'main' panicked at src/xgboost/xgb_regressor.rs:323:21:
attempt to subtract with overflow
```

Expected: either a `Failed::because(FailedError::ParametersError, ...)` explaining that the subsample ratio is too small for this many rows, or a sample size clamped to at least 1.

Note this is debug-profile behavior. In release the overflow check is off, so the wrapped range would index out of bounds instead.

smartcore 0.6.11, rustc 1.95.0, on `main` at 6e285b3.

I have a fix and tests ready and will open a PR against this issue if the approach above sounds right.

teddytennant:
## Description

`XGRegressor::fit` panics with `attempt to subtract with overflow` when `subsample` is small enough that the sampled index set comes out empty. `sample_without_replacement` sizes the sample as `(population_size as f64 * subsample_ratio) as usize`, which floors to 0. Three rows at `subsample = 0.3` does it, and so does a single row at any ratio below 1.0. `find_best_split` then runs `0..sorted_idxs.len() - 1` on an empty slice.

The guard in `fit` only rejects `subsample > 1.0` and `subsample <= 0.0`, so this input gets through.

Fixes #444

## Current behaviour

```
thread 'main' panicked at src/xgboost/xgb_regressor.rs:323:21:
attempt to subtract with overflow
```

## New expected behaviour

The sample keeps a minimum of one row, so the fit succeeds and returns a usable model. This is what scikit-learn does for its own `subsample` parameter (`max(1, int(subsample * n_samples))` in `sklearn/ensemble/_gb.py`).

I went this way rather than returning `Failed::because(FailedError::ParametersError, ...)` because `subsample` is a regularisation hyperparameter, not a statement about the data. The parameter is valid on its own; only the pair `(subsample, n_samples)` is degenerate. Erroring would mean a `subsample` that works on the full dataset starts failing inside `cross_validate` as k-fold shrinks each fold, which is a surprising place to hit it.

Sample sizes of one row or more are byte-identical to before. `.max(1)` only engages when the floor is 0, and `shuffle` consumes the same RNG state either way.

## Tests

Four tests, in the module's own `mod tests`. Three fail before this change and pass after: 3 rows at 0.3, 1 row at 0.9, and a table over the boundary. The fourth is the regression guard, checking that ratios which already yield one or more rows are untouched: `(10, 0.8) -> 8`, `(100, 0.8) -> 80`, indices unique and in range. `(2, 0.4)` and `(2, 0.5)` sit either side of the floor boundary, so the pair pins the clamp rather than just testing small numbers.

Lib tests go 428 to 432.

## Change logs

`CHANGELOG.md` gets a `### Fixed` bullet under a new `## [0.6.12]`, and `Cargo.toml` is bumped to match, following #443.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,10 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [0.6.12]
+### Fixed
+- `xgboost/xgb_regressor.rs`: `XGRegressor::fit` no longer panics with `attempt to subtract with overflow` when `subsample` is less than 1.0 on a small dataset (#444). `floor(n_samples * subsample)` truncates to 0 rows. This occurs with 3 rows at a ratio of 0.3, or with 1 row at any ratio below 1.0. The tree fit then read `sorted_idxs.len() - 1` on an empty index set. The sample size now keeps a minimum of one row, as scikit-learn does for its own `subsample` parameter. Sample sizes of one row or more are unchanged.
+
 ## [0.6.11]
 ### Fixed
 - `algorithm/neighbour/cosinepair.rs`: `CosinePair::query_row_top_k` now returns exact nearest neighbours whenever `approximate` is `false` (the default). Previously the query always sampled only `top_k` evenly strided candidate rows without documentation, and the bounded candidate heap evicted its closest entry, so the method could return the farthest of the sampled rows (#442). Strided sampling is now gated behind `CosinePairParameters { approximate: true, .. }` and is documented as approximate.
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -2,7 +2,7 @@
 name = "smartcore"
 description = "Machine Learning in Rust."
 homepage = "https://smartcorelib.github.io/"
-version = "0.6.11"
+version = "0.6.12"
 authors = ["smartcore Developers"]
 edition = "2024"
 rust-version = "1.85"
--- a/src/xgboost/xgb_regressor.rs
+++ b/src/xgboost/xgb_regressor.rs
@@ -491,6 +491,8 @@ impl XGRegressorParameters {
     /// Sets the fraction of samples to be used for fitting individual base learners.
     ///
     /// A value of less than 1.0 introduces randomness and helps prevent overfitting.
+    /// The value must be in the range (0, 1]. Each tree gets `floor(n_samples * subsample)`
+    /// rows, but a minimum of one row.
     pub fn with_subsample(mut self, subsample: f64) -> Self {
         self.subsample = subsample;
         self
@@ -599,6 +601,9 @@ impl<TX: Number + PartialOrd, TY: Number, X: Array2<TX>, Y: Array1<TY>> XGRegres
     }
 
     /// Creates a random sample of indices without replacement.
+    ///
+    /// The sample holds at least one index when the population is not empty. The tree fit
+    /// needs a minimum of one row, thus the sample size cannot be zero.
     fn sample_without_replacement(
         population_size: usize,
         subsample_ratio: f64,
@@ -606,7 +611,10 @@ impl<TX: Number + PartialOrd, TY: Number, X: Array2<TX>, Y: Array1<TY>> XGRegres
     ) -> Vec<usize> {
         let mut indices: Vec<usize> = (0..population_size).collect();
         indices.shuffle(rng);
-        indices.truncate((population_size as f64 * subsample_ratio) as usize);
+        // `population_size * subsample_ratio` truncates to 0 for a small population, e.g. 3 rows
+        // at a ratio of 0.3. Keep one row in that case, as scikit-learn does for its own
+        // `subsample` parameter (#444).
+        indices.truncate(((population_size as f64 * subsample_ratio) as usize).max(1));
         indices
     }
 }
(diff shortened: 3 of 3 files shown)


Mec-iS:
thanks 🙏 i will look into this asap

Mec-iS:
## Review: ✅ Approve with minor notes

The fix is correct, minimal, and safe. The one-line change in `sample_without_replacement` is the right approach and matches the established scikit-learn convention.

---

### The Core Fix

```rust
// Before
indices.truncate((population_size as f64 * subsample_ratio) as usize);

// After
indices.truncate(((population_size as f64 * subsample_ratio) as usize).max(1));
```

This is correct and complete. The `.max(1)` only engages on the underflow case (`floor → 0`), so all existing behaviour for sample sizes ≥ 1 is byte-identical. The panic in `find_best_split` at `0..sorted_idxs.len() - 1` on an empty slice is fully prevented. The rationale for clamping rather than erroring (cross-validation folds shrinking the dataset) is well-argued and consistent with scikit-learn's `max(1, int(subsample * n_samples))`.

---

### Test Coverage

The four tests cover the important boundaries well:

- `test_subsample_smaller_than_one_sample_does_not_panic` — end-to-end fit/predict with 3 rows at 0.3 (the original bug)
- `test_subsample_on_single_row_does_not_panic` — single-row edge case
- `test_sample_without_replacement_clamps_to_one_row` — unit test covering multiple `(population, ratio)` pairs
- `test_sample_without_replacement_keeps_the_ratio_above_one_row` — regression guard ensuring `.max(1)` doesn't inflate normal samples

One gap worth noting: there is no test for `population_size == 0`. Currently `sample_without_replacement(0, any_ratio, rng)` will call `truncate(1)` on an empty `Vec`, which is a safe no-op in Rust and returns `[]`. This is probably fine since `fit` validates `n_samples > 0` upstream, but a short comment or `debug_assert!(population_size > 0)` would make the invariant explicit.

---

### Minor Suggestions

- **`with_subsample` docstring**: the new line says *"a minimum of one row"* — worth adding a parenthetical like `(provided the dataset is non-empty)` for precision.
- **`sample_without_replacement` docstring**: same clarification would help here.
- **`CHANGELOG.md`**: the entry is accurate but the last two sentences read more like a commit message than a changelog bullet — consider trimming to just the user-visible behaviour.

None of these block merging. Great fix overall — clean, targeted, and well-tested.

Mec-iS:
## Second Pass Review

Having looked more carefully at the diff, a few additional observations beyond my first comment.

---

### Correctness of the fix in context of `find_best_split`

The panic originates at `0..sorted_idxs.len() - 1` on an empty slice. It is worth noting this is a `usize` underflow, not a bounds panic — `0 - 1` wraps to `usize::MAX` in debug builds and panics explicitly; in release it silently produces an enormous range. The `.max(1)` guard fully eliminates the empty-slice case, so both paths are covered. ✅

### The `shuffle`-before-`truncate` ordering

The PR notes that `shuffle` consumes the same RNG state regardless of the clamp. This is true and important: the fix does **not** change the RNG sequence for any call where the sample size was already ≥ 1, so existing serialised/reproducible models are unaffected. ✅

### Potential issue: `subsample_ratio` not clamped to `[0, 1]` inside `sample_without_replacement`

The function signature accepts any `f64`. If `subsample_ratio > 1.0` somehow reaches it (e.g. via a future refactor that bypasses `fit`'s guard), `(population_size as f64 * subsample_ratio) as usize` silently exceeds `population_size`, and `truncate` on a shorter vec is a no-op — so you get all rows rather than erroring. This is benign today, but a `debug_assert!(subsample_ratio > 0.0 && subsample_ratio <= 1.0)` at the top of the function would make the contract explicit at zero runtime cost in release builds.

### Missing blank line between tests

```rust
        assert!(predictions[0].is_finite());
    }
    #[test]   // ← no blank line before this attribute
    fn test_sample_without_replacement_clamps_to_one_row() {
```

Minor style nit — all other test functions in this module have a blank line separating them. `rustfmt` won't flag it but it breaks the visual rhythm.

### `test_subsample_on_single_row_does_not_panic` — consider asserting the predicted value

With a single-row dataset and 5 estimators each forced to train on that one row, the model should perfectly memorise the target. Adding `assert!((predictions[0] - 5.0_f64).abs() < 1e-6)` (or similar tolerance) would make the test document the expected numerical behaviour, not just the absence of a panic.

---

None of these are blockers. The core change is sound and I'd be happy to see this merged as-is. The `debug_assert` suggestion is the only one I'd consider worth a follow-up issue if not addressed here.

Mec-iS (approved):

PR #456, opened 2026-08-25
Add LDA for supervised dimensionality reduction (#136)

### LINKED ISSUE #136: Add Linear Discriminant Analysis (LDA) support

ChrisJr404:
Fixes #136

### Checklist
- [x] My branch is up-to-date with main branch.

### Current behaviour
`decomposition` only offers unsupervised reduction (`PCA`, `SVD`). There is no way to reduce the features using the class labels, so the directions that best separate the classes are not available.

### New expected behaviour
Adds `LDA`, linear discriminant analysis for supervised dimensionality reduction, next to `PCA` in the same module. `LDA::fit(x, y, params)` learns the discriminant directions and `transform` projects onto them, so it plugs into the existing `Transformer` interface.

Implementation notes:
- It solves the generalized eigenproblem `Sb w = lambda Sw w` by whitening with the within-class scatter matrix. `Sw` is factored once with a symmetric eigen decomposition to build `Sw^(-1/2)`, then `Sw^(-1/2) Sb Sw^(-1/2)` is symmetric and a second symmetric eigen decomposition gives the directions. That keeps everything on the existing symmetric `evd` path, no external numeric crates and no unsafe.
- By default it keeps `min(n_classes - 1, n_features)` components, matching the number of useful directions. `with_n_components` lets you ask for fewer.
- Singular within-class scatter (for instance more features than samples per class) returns `Failed` instead of producing garbage.
- The directions and the projection match scikit-learn 1.9.0 `LinearDiscriminantAnalysis(solver="eigen")` up to the usual per-axis sign, and the known-answer test pins those reference values (compared on magnitudes, the way the PCA tests do).

I scoped this to the dimensionality-reduction side to keep it close to `PCA`. Happy to follow up with an LDA classifier (`predict`) in a separate PR if that is wanted.

#### Added
- `decomposition::lda::LDA` with `LDAParameters` / `LDASearchParameters`, a `Transformer` implementation, a `scalings` getter, doctest and unit tests.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,10 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [0.6.13]
+### Added
+- `decomposition/lda.rs`: `LDA`, linear discriminant analysis for supervised dimensionality reduction (#136). It projects the data onto the directions that best separate the classes, keeping `min(n_classes - 1, n_features)` components by default, and implements the `Transformer` interface next to `PCA`. Directions match scikit-learn's `LinearDiscriminantAnalysis(solver="eigen")` up to sign.
+
 ## [0.6.12]
 ### Fixed
 - `xgboost/xgb_regressor.rs`: `XGRegressor::fit` no longer panics when `subsample` is less than 1.0 on a small dataset (#444). The sample for each tree now keeps a minimum of one row, as scikit-learn does for its own `subsample` parameter. Sample sizes of one row or more are unchanged.
--- a/src/decomposition/lda.rs
+++ b/src/decomposition/lda.rs
@@ -1,0 +1,518 @@
+//! # Linear Discriminant Analysis
+//!
+//! Linear discriminant analysis (LDA) finds the linear combinations of the features that best separate two or more classes.
+//! Unlike [PCA](../pca/index.html), which is unsupervised and keeps the directions of largest variance, LDA is supervised: it uses the class
+//! labels \\(y\\) to look for the directions that maximise the ratio of between-class scatter to within-class scatter (the Rayleigh quotient).
+//! The result is a projection of the data onto at most \\(C - 1\\) axes, where \\(C\\) is the number of classes, that pulls the classes apart.
+//!
+//! LDA is often used to reduce the number of features before a classification step and for data visualization.
+//! Each discriminant direction \\(w\\) is a generalized eigenvector of the pair \\((S_B, S_W)\\), the between-class and within-class scatter matrices:
+//!
+//! \\[S_B w = \lambda S_W w\\]
+//!
+//! The directions are kept in descending order of \\(\lambda\\), so the first component separates the classes the most.
+//!
+//! LDA assumes the classes share the same covariance. It needs the within-class scatter matrix to be invertible, so keep the number of
+//! samples per class larger than the number of features.
+//!
+//! Example:
+//! ```
+//! use smartcore::linalg::basic::matrix::DenseMatrix;
+//! use smartcore::decomposition::lda::*;
+//!
+//! // Three well separated classes, three samples each
+//! let x = DenseMatrix::from_2d_array(&[
+//!                     &[4.0, 2.0, 0.6],
+//!                     &[4.2, 2.1, 0.5],
+//!                     &[3.9, 1.9, 0.7],
+//!                     &[6.0, 3.0, 4.5],
+//!                     &[6.2, 2.9, 4.6],
+//!                     &[5.8, 3.1, 4.4],
+//!                     &[7.5, 3.6, 6.1],
+//!                     &[7.7, 3.5, 6.0],
+//!                     &[7.3, 3.7, 6.2],
+//!                     ]).unwrap();
+//! let y = vec![0, 0, 0, 1, 1, 1, 2, 2, 2];
+//!
+//! let lda = LDA::fit(&x, &y, LDAParameters::default()).unwrap(); // keep C - 1 = 2 components
+//!
+//! let projected = lda.transform(&x).unwrap();
+//! ```
+//!
+//! ## References:
+//! * ["An Introduction to Statistical Learning", James G., Witten D., Hastie T., Tibshirani R., 4.4 Linear Discriminant Analysis](http://faculty.marshall.usc.edu/gareth-james/ISL/)
+//! * ["Pattern Classification", Duda R.O., Hart P.E., Stork D.G., 2nd ed., 3.8.3 Multiple Discriminant Analysis](https://www.wiley.com/en-us/Pattern+Classification%2C+2nd+Edition-p-9780471056690)
+//!
+//! <script src="https://polyfill.io/v3/polyfill.min.js?features=es6"></script>
+//! <script id="MathJax-script" async src="https://cdn.jsdelivr.net/npm/mathjax@3/es5/tex-mml-chtml.js"></script>
+use std::cmp::Ordering;
+use std::fmt::Debug;
+
+#[cfg(feature = "serde")]
+use serde::{Deserialize, Serialize};
+
+use crate::api::Transformer;
+use crate::error::Failed;
+use crate::linalg::basic::arrays::{Array1, Array2};
+use crate::linalg::traits::evd::EVDDecomposable;
+use crate::numbers::basenum::Number;
+use crate::numbers::realnum::RealNumber;
+
+/// Linear discriminant analysis
+#[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
+#[derive(Debug)]
+pub struct LDA<T: Number + RealNumber, X: Array2<T> + EVDDecomposable<T>> {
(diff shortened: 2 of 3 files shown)


Mec-iS:
thanks. please consider adding tests as suggested by the coverage bot

Mec-iS:
also: v0.6.13 has already been published.
New one is v0.6.14

Mec-iS:
Thanks for this contribution — the implementation is clean and well-structured. Here are my notes from a full diff review.

---

## ✅ What works well

- The whitening approach (`Sw^(-1/2) Sb Sw^(-1/2)`) is mathematically correct and a good fit for the existing symmetric `evd` path. No new dependencies needed.
- `min(n_classes - 1, n_features)` default for `n_components` is correct and matches sklearn.
- The singularity guard (`li <= T::epsilon()`) is sensible, though see note below.
- The known-answer test against sklearn 1.9.0 with sign-agnostic magnitude comparison is the right approach, consistent with the PCA tests.
- `serde` round-trip test is appreciated.
- Doc-comment math and references are thorough.

---

## 🐛 Bugs / Correctness

**1. Singularity threshold is too tight**

```rust
if li <= T::epsilon() {
```

`T::epsilon()` is the machine epsilon (~2.2e-16 for f64), which is too small a floor — numerically near-singular matrices will pass this check and produce garbage whitening. A practical threshold like `li < T::from(1e-10).unwrap()` (or relative to the largest eigenvalue) would be safer. sklearn uses `1e-4` relative to the largest singular value.

**2. CHANGELOG version mismatch**

The CHANGELOG entry is filed under `[0.6.13]`, but the maintainer noted that `v0.6.13` has already been published. This entry should be bumped to `[0.6.14]`.

---

## ⚠️ Missing coverage (33 uncovered lines per codecov)

The bot flagged 72.5% patch coverage. The paths most likely uncovered are:

- `n_components = Some(c)` validation branch (the `c < 1 || c > max_components` error arm)
- The `n != y.shape()` dimension mismatch error path
- The `transform` shape-mismatch error path
- The `whitening` / intermediate `evd` failure path

Suggested additions:
```rust
#[test]
fn mismatched_x_y_is_rejected() {
    let (x, _) = three_class_data();
    let y_short = vec![0i32; x.shape().0 - 1];
    assert!(LDA::fit(&x, &y_short, LDAParameters::default()).is_err());
}

#[test]
fn zero_n_components_is_rejected() {
    let (x, y) = three_class_data();
    assert!(LDA::fit(&x, &y, LDAParameters::default().with_n_components(0)).is_err());
}

#[test]
fn transform_wrong_features_is_rejected() {
    let (x, y) = three_class_data();
    let lda = LDA::fit(&x, &y, LDAParameters::default()).unwrap();
    let bad = DenseMatrix::<f64>::zeros(3, 2); // wrong n_features
    assert!(lda.transform(&bad).is_err());
}
```

---

## 🔧 Minor / Style

**3. `PartialEq` uses `T::epsilon()` for floating-point equality**

Structural equality comparisons across eigenvectors are sensitive to floating-point order. Consider documenting this limitation or using a relative tolerance, consistent with how `approx::relative_eq!` is used in tests.

**4. `scalings` field name shadows the getter**

The private field `scalings` and the public method `scalings()` have the same identifier. This compiles fine in Rust (method vs field), but it may confuse readers. Consider renaming the field to `scalings_` or `projection_matrix` to make the distinction explicit — this is also the convention used in the PCA struct.

**5. `pub mod lda;` ordering in `mod.rs`**

The new `pub mod lda;` line is placed *before* the `pca` entry with no blank line. The comment above it describes LDA but the `/// PCA is a popular approach...` doc comment for the pca module is now visually detached. A blank line between the two module declarations would keep the module-level docs tidy:

```rust
/// LDA is a supervised approach ...
pub mod lda;

/// PCA is a popular approach ...
pub mod pca;
pub mod svd;
```

**6. `polyfill.io` script tag in doc comment**

```rust
//! <script src="https://polyfill.io/v3/polyfill.min.js?features=es6"></script>
```

`polyfill.io` was acquired and its CDN has had documented supply-chain concerns. The PCA module uses MathJax directly without polyfill.io — recommend removing this line for consistency and safety.

---

## 💡 Follow-up suggestions (non-blocking)

- A `predict` method (nearest-centroid classification on the projected space) would complete the LDA story and aligns with the issue's original request. Great to scope that as a separate PR as proposed.
- Consider exposing `explained_variance_ratio_` (each eigenvalue divided by the sum) analogous to `PCA`, useful for scree plots.
- The `LDASearchParameters` struct and iterator are complete but the default only iterates `[None]`. A short note in the docs clarifying how to construct a non-trivial grid would help users.

---

Overall this is a solid first pass. The core math is right, the API is idiomatic, and the tests pin reference values well. Addressing the CHANGELOG version, the singularity threshold, and the missing test coverage are the main blockers before merge.

Mec-iS:
Hi @ChrisJr404 — thanks for this contribution, the core math and API shape are solid. I reviewed the full diff and have the following notes (ordered roughly by severity).

---

## 🐛 Bugs / Correctness

### 1. Singularity threshold is too tight

```rust
// current
if li <= T::epsilon() {
```

`T::epsilon()` is the machine epsilon (~2.2e-16 for f64), which is far too tight. Numerically near-singular `Sw` matrices will sail right past this guard and produce garbage whitening vectors. A safer floor is a small absolute threshold *or* a value relative to the largest eigenvalue:

```rust
let l_max = l.iter().cloned().fold(T::zero(), |a, b| if b > a { b } else { a });
let tol = T::from(1e-10).unwrap().max(T::from(1e-6).unwrap() * l_max);
for &li in &l {
    if li <= tol {
        return Err(Failed::fit(
            "Within-class scatter matrix is singular …",
        ));
    }
}
```

scikit-learn uses `1e-4` relative to the largest singular value for the same purpose.

---

### 2. CHANGELOG entry is under the wrong version

`v0.6.13` has already been published (as noted by the maintainer). This entry should be moved to `## [0.6.14]`.

---

## ⚠️ Missing test coverage (33 lines, ~27.5 % of the patch)

The uncovered paths are almost certainly:

| Missing branch | Suggested test |
|---|---|
| `n_components = Some(c)` validation (`c < 1`) | `zero_n_components_is_rejected` |
| `n != y.shape()` length mismatch | `mismatched_x_y_is_rejected` |
| `transform` wrong-feature-count path | `transform_wrong_features_is_rejected` |
| `evd` failure / intermediate whitening failure | hard to trigger directly; partial coverage OK |

Concrete additions you could drop into the existing `tests` module:

```rust
#[test]
fn mismatched_x_y_is_rejected() {
    let (x, _) = three_class_data();
    let y_short = vec![0i32; x.shape().0 - 1];
    assert!(LDA::fit(&x, &y_short, LDAParameters::default()).is_err());
}

#[test]
fn zero_n_components_is_rejected() {
    let (x, y) = three_class_data();
    assert!(LDA::fit(&x, &y, LDAParameters::default().with_n_components(0)).is_err());
}

#[test]
fn transform_wrong_features_is_rejected() {
    let (x, y) = three_class_data();
    let lda = LDA::fit(&x, &y, LDAParameters::default()).unwrap();
    let bad = DenseMatrix::<f64>::zeros(3, 2); // wrong n_features
    assert!(lda.transform(&bad).is_err());
}
```

---

## 🔧 Minor / Style

### 3. `PartialEq` uses `T::epsilon()` for floating-point equality

Same issue as point 1: comparing eigenvectors with `T::epsilon()` will be brittle in practice. Consider using a small fixed tolerance (e.g. `T::from(1e-10).unwrap()`) or documenting the limitation explicitly, consistent with how `approx::relative_eq!` is used in the tests.

### 4. Private field `scalings` shadows the public getter `scalings()`

Rust resolves this without ambiguity, but the dual use of the identifier confuses readers. The PCA struct uses a distinct field name (`components`) from its public accessors. Consider renaming the field to `projection_matrix` (or `scalings_` at minimum):

```rust
pub struct LDA<T, X> {
    projection_matrix: X,   // was `scalings`
    eigenvalues: Vec<T>,
    n_features: usize,
}
```

### 5. Missing blank line between `pub mod lda` and `pub mod pca` in `mod.rs`

The doc comment for the `pca` module is currently visually attached to the `lda` declaration. A blank line restores the original formatting:

```rust
/// LDA is a supervised approach …
pub mod lda;

/// PCA is a popular approach …
pub mod pca;
pub mod svd;
```

### 6. `polyfill.io` script tag in doc comment

```rust
//! <script src="https://polyfill.io/v3/polyfill.min.js?features=es6"></script>
```

`polyfill.io` was acquired in 2024 and its CDN has had documented supply-chain concerns. The PCA module already uses MathJax directly without polyfill.io. Recommend removing this line entirely for consistency and safety.

---

## 💡 Non-blocking suggestions

- **`explained_variance_ratio_`** — exposing each eigenvalue divided by the total (analogous to PCA) would make scree plots trivial for users.
- **`LDASearchParameters` docs** — the default grid only iterates `[None]`. A short doc note explaining how to construct a non-trivial hyperparameter grid would help users adopting this with the existing grid-search infra.
- **`predict` (nearest-centroid)** — great call scoping that to a follow-up PR; it would complete the supervised story opened in #136.

---

Overall this is a clean first pass. The whitening algebra is correct, the sklearn reference test is the right approach, and the API fits idiomatically into the existing `decomposition` module. Addressing the **CHANGELOG version**, **singularity threshold**, and **missing test coverage** are the main items before merge.

